PR #78, opened 2016-11-06
fix: memberList Method have no 'name' argument defined ,it'will cause a undefined error.

JasLin:
```
ERR! processRooms ReferenceError: name is not defined
ERR! processRooms     at Room.memberList (/home/jas/projects/learning/wechat/wechaty-ts/src/room.ts:367:39)
```


zixia:
Thanks!

PR #87, opened 2016-11-12
fill in the TBW block

lijiarui:
Contact,FriendRequest,Message,Room,RoomEvent


zixia:
nice.

could you also add an one line description to each API?

thanks!

lijiarui:
done.

zixia:
nice. thanks!

zixia:
I found there has some linking issue in your doc after I reformated them.

The link to the anchor at API Wiki not work as expected. They can only link to Wiki, but not the right anchor.

Could you fix that?

PR #109, opened 2016-11-28
Qrcodeimg

lijiarui:
qrcode img png


zixia:
do not include room.ts
do not include unnecessary image

Issue #173, opened 2016-12-30
room.member() cannot find contact when contact set whose alias in the room


lijiarui:
### Run `npm run doctor` or `wechaty run doctor`(for docker user), paste output here 
```shell
1. Wechaty version: #git[d6b69ab #155]
2. Darwin x64 version 14.5.0 memory 2398/16384 MB
3. Docker: false
4. Node version: v7.0.0
```

### Expected behavior

Use room.member() to find contact by name in a room. 

### Actual behavior

Cannot find contact by name when the contact change whose alias in the room.

### Steps to reproduce the behavior (and fixes, if any)

code as follows:

```
async function RoomTest(){
  let room = await Room.find({topic:"wetest"})
  await room.ready()
  console.log("member1: " + room.member("wuli舞哩客服"))
  console.log("member2: " + room.member("桔小秘"))
}
```
log as follows:
```
member1: wuli舞哩客服
member2: null
```
contact `"桔小秘"` has changed her alias in the room, so room.member() return null.

I found your annotation in `room.ts` says：
```
/**
   * NickName / DisplayName / RemarkName of member
   */
public member(name: string): Contact | null {
   // your function
}
```
I thought maybe you mean we can find room member by nickname、displayname or remark.

while, you use following function to find member

```
const idList = Object.keys(nickMap)
                          .filter(k => nickMap[k] === name)
```

nickMap is a map with id and displayName, log as follows:

```
SILL Room member() check nickMap: {"@bba74f144ab75df78e9cd3bbe9f79937":"李佳芮","@e40ab116adf563b28e0a8efdfb737b5bdfef69fad6dd83a02f087b827d8238b7":"哈哈","@c1d97b69778da9a480e790b15666fcbe3dfb0e185a5cbf3a07a813b21579dd68":"wuli舞哩客服"}
```

so, I thought may be you should refract nickMap to fix this bug?

Could you help me to fix it? thanks

zixia:
Not sure about your issue, but I know there will be some issue because the Tencent had changed the way of show names in the room for many times.

You can just send a PR(with the unit test), then I'll be able to dive into it and understand which part you are going to fix.

lijiarui:
I mean function `room.member()` cannot work well when the contact set its' remark, including in group and in contact list.

lijiarui:
Contact's name was shown by the following order:   

remark set by bot  > remark set by itself in the group > nickName

it means:

[...]

Issue #104, opened 2016-11-24
room.member() cannot get member when bot set remark for friend


lijiarui:
### Run `npm run doctor` or `wechaty run doctor`(for docker user), paste output here 

```shell
> wechaty@0.6.22 doctor /wechaty
> ts-node bin/doctor


#### Wechaty Doctor

1. Wechaty version: 0.6.22
2. Linux x64 version 4.4.27-moby memory 1342/1997 MB
3. Docker: true
4. Node version: v7.1.0
```

### Steps to reproduce the behavior (and fixes, if any)

When the bot modify it's friend's nickname [Alice] to remarkname [AliceRevise],  the other user in the room want to use [@Alice] to let the bot know he wants to do something to Alice, but when use function `room.member('Alice')` , the bot cannot find Alice by Nickname, when change to @AliceRevise, using function room.member('AliceRevise'),then the bot can find Alice.

I found function `room.member()` try to detect nickMap in the room to find the contact, when the bot modifies nickname, you assign RemarkName to nickMap, so the bot cannot use room.member() find the user.

I guess it's a bug, you should assign NickName not RemarkName to nickMap.

### Paste the full output logs here with `WECHATY_LOG=silly` set

```
SILL Room member() check nickMap: {"@f617961475d7d90e6f22d24caff884ef":"李佳芮","@76185926dff01e5b6210944001d9cd0c":"AliceRevise","@48614ff69405db58046563bdb7f1385012f13e64a6b588b22b7c888ea52e0f0d":"kz社区","@152c1e75119f393005532ea3131ad2b2":"Love-U-Unconditionally","@c536eee2244de14c707bdde0c8f11d658ffe3f5562953f72ad34ced3a7e1c054":"桔子"}
```

console.log(room) result as follows:

```
Room {
  domain: null,
  _events: {},
  _eventsCount: 0,
  _maxListeners: undefined,
  id: '@@7bcae57dcded44e764b90f62f4a2dc5ae08d30a34a9488aa96bf2ed3e3ee6feb',
  rawObj:
   { Alias: '',
     AppAccountFlag: 0,
     AttrStatus: 0,
     ChatRoomId: 0,
     City: '',
     ContactFlag: 0,
     DisplayName: '',
     EncryChatRoomId: '@34633f0f3a40068585fa11489b6442b9',
[...]

Issue #64, opened 2016-10-26
当用户昵称中含有表情时，无法出发room-join 事件


lijiarui:
#### Wechaty Doctor
1. Wechaty version: #git[32234c2 fix sloc file glob bug in win32]
2. Darwin x64 version 15.6.0 memory 24/4096 MB
3. Docker: false
### Expected behavior

当用户昵称中含有表情的时候，该修改群名称，可以触发room-topic 事件。并将群名称改回来。
用户昵称为：李佳芮😂
代码如下：

.on('room-topic',  async function(this, room, topic, oldTopic, changer) {
  try {
    log.info('Bot', 'EVENT: room-topic - Room %s change topic  to %s by member %s'
                  , oldTopic
                  , topic
                  , changer.name()
                )

```
const user = bot.user()
const user_id = user.id
if (!user) {
  throw new Error('no user')
}

if(user_id !== changer.id){
  room.say("@" + changer.name() + "  " + RoomNoChangeTopic)  
  room.topic(String(oldTopic))   
}
```

  } catch (e) {
    log.error('Bot', 'room-topic event exception: %s', e.stack)
  }

  return
})
### Actual behavior

当用户昵称中含有表情的时候，修改昵称后。无法出发room-join事件
### Paste the full output logs here with `WECHATY_LOG=silly` set

当用户昵称含有表情时候的contact日志：
Contact {
  id: '@0578030b97ba8e15b0a489f1c5cf9326',
  rawObj: 
   { '$$hashKey': '01X',
     Alias: 'ruirui_0914',
     AppAccountFlag: 0,
     AttrStatus: 100455,
     ChatRoomId: 0,
     City: '海淀',
     ContactFlag: 3,
     DisplayName: '',
     EncryChatRoomId: '0',
     HeadImgUrl: '/cgi-bin/mmwebwx-bin/webwxgeticon?seq=635310315&username=@0578030b97ba8e15b0a489f1c5cf9326&skey=',
     HideInputBarFlag: 0,
     KeyWord: 'qq5',
     MMFromBatchGet: true,
     MMFromBatchget: true,
     MMOrderSymbol: 'LIJIARUISPANCLASSEMOJIEMOJI1F602SPAN',
     MemberCount: 0,
     MemberList: [],
     NickName: '李佳芮<img class="emoji emoji1f602" text="_web" src="/zh_CN/htmledition/v2/images/spacer.gif" />',
     OwnerUin: 0,
[...]

PR #211, opened 2017-01-17
fix_function_room.member_#173

lijiarui:
#173 
#104 

1. change parseNickMap, add remark value
2. add contact.ready() to get full contact, and get contact's remark


lijiarui:
Sorry， I found I fix what I need, but bring in other bug, so please wait until I fix all

lijiarui:
this commit can help to find member in room by nick/display/remark

1. add following type:

```
type MemberQueryFilter = {
  nick?: string
  remark?: string
  display?: string
}
```

2. add following Map:
```
nickMap,
remarkMap,
displayMap,
```

3.  add following Functions：
```
parseNickMap()
parseRemarkMap()
parseDisplayMap()
```

4. change the parameter in function `room.member(name: string)` from `public member(quaryArg: MemberQueryFilter)`   
we can find member by nick, display, remark, code as follows:

```
room.member({nick: '李佳芮'})
room.member({remark: '哈哈哈'})
room.member({display: '土'})
```

lijiarui:
@zixia 
I have tested it, it works well with the following function:
```
async function RoomTest() {
  let room = await Room.find({topic: 'test'})
  await room.ready()
  console.log('member1: ===========')
  console.log(room.member({nick: '李佳芮'}))
  console.log('member2: ')
  console.log(room.member({remark: '哈哈哈'}))
  console.log('member3: ')
  console.log(room.member({display: '土'}))
}
```

I didn't modify getContact and related functions, but [test](https://ci.appveyor.com/project/Wechaty/wechaty/build/270) shows:

> test » room » Room smoking test Error: Cannot read property 'getContact' of undefined

Could you help me to fix? Thanks.

src/room.ts
```diff
@@ -209,21 +221,53 @@ export class Room extends EventEmitter implements Sayable {
     return rawMemberList.map(m => Contact.load(m.UserName))
   }
 
-  private parseNickMap(memberList: RoomRawMember[]): Map<string, string> {
+  private async parseNickMap(memberList: RoomRawMember[]): Promise<Map<string, string>> {
```
zixia: Do not use `async` here. use `readyMembers()` in my review suggestion instead.

src/room.ts
```diff
@@ -182,14 +190,16 @@ export class Room extends EventEmitter implements Sayable {
 
   public get(prop): string { return (this.obj && this.obj[prop]) || (this.dirtyObj && this.dirtyObj[prop]) }
 
-  private parse(rawObj: RoomRawObj): RoomObj | null {
+  private async parse(rawObj: RoomRawObj): Promise<RoomObj | null> {
```
zixia: Keep `parse()` a sync function is better.

src/room.ts
```diff
@@ -114,7 +122,7 @@ export class Room extends EventEmitter implements Sayable {
       const data = await contactGetter(this.id)
       log.silly('Room', `contactGetter(${this.id}) resolved`)
       this.rawObj = data
-      this.obj    = this.parse(data)
+      this.obj    = await this.parse(data)
```
lijiarui: I guess you typo here..
> Also parseNickMap() parseRemarkMap() parseDisplayName() should not be async function too.

you mean parseNickMap() parseRemarkMap() parseDisplayName() should be sync function?

src/puppet-web/firer.ts
```diff
@@ -207,7 +207,7 @@ async function checkRoomJoin(m: Message): Promise<void> {
         const loaded = inviteeContactList[i] instanceof Contact
 
         if (!loaded) {
-          let c = room.member(inviteeList[i])
+          let c = room.member({remark: inviteeList[i]}) || room.member({nick: inviteeList[i]})
```
lijiarui: > We can keep room.member(inviteeList[i]) here, after you follow the review for member() parameters suggestion.

I don't agree with this suggestion, because of the following reason:

We keep function `room.member()` work, but it should work only with the function find member by nickName.    
We cannot use it both find remark and nickName, because one can set a remark same as nick, then it would be confused. 

So, what do you think about it?

src/puppet-web/firer.ts
```diff
@@ -234,7 +234,7 @@ async function checkRoomJoin(m: Message): Promise<void> {
       }
 
       if (!inviterContact) {
-        inviterContact = room.member(inviter)
+        inviterContact = room.member({remark: inviter}) || room.member({nick: inviter})
```
lijiarui: > We can keep room.member(inviteeList[i]) here, after you follow the review for member() parameters suggestion.

I don't agree with this suggestion, because of the following reason:

We keep function `room.member()` work, but it should work only with the function find member by nickName.    
We cannot use it both find remark and nickName, because one can set a remark same as nick, then it would be confused. 

So, what do you think about it?

src/puppet-web/firer.ts
```diff
@@ -316,7 +316,7 @@ async function checkRoomLeave(m: Message): Promise<void> {
   /**
    * FIXME: leaver maybe is a list
    */
-  let leaverContact = room.member(leaver)
+  let leaverContact = room.member({remark: leaver}) || room.member({nick: leaver})
```
lijiarui: > We can keep room.member(inviteeList[i]) here, after you follow the review for member() parameters suggestion.

I don't agree with this suggestion, because of the following reason:

We keep function `room.member()` work, but it should work only with the function find member by nickName.    
We cannot use it both find remark and nickName, because one can set a remark same as nick, then it would be confused. 

So, what do you think about it?

src/puppet-web/firer.ts
```diff
@@ -367,7 +366,7 @@ async function checkRoomTopic(m: Message): Promise<void> {
   if (/^You$/.test(changer) || /^你$/.test(changer)) {
     changerContact = Contact.load(this.userId)
   } else {
-    changerContact = room.member(changer)
+    changerContact = room.member({remark: changer}) || room.member({nick: changer})
```
lijiarui: > We can keep room.member(inviteeList[i]) here, after you follow the review for member() parameters suggestion.

I don't agree with this suggestion, because of the following reason:

We keep function `room.member()` work, but it should work only with the function find member by nickName.    
We cannot use it both find remark and nickName, because one can set a remark same as nick, then it would be confused. 

So, what do you think about it?

lijiarui:
@zixia  done for what you request

src/puppet-web/firer.ts
zixia: Why you copy/paste the same answer four times here?

I comment more than one times because I want to point out the position for you where needs to change.

What you were arguing about is not wrong, but it's not entirely right. Remember always to KISS(Keep It Simple, Stupid). 

The most scenario is we just want to get a member by name(whatever remark/nick/display). I do not agree with you that "it should work only with the function find the member by nickname", instead it should return a contact for any name.  And If "one set a remark same as nick", then there's the time to use the new QueryFilter as the parameter.

So just change it to what I said.

src/room.ts
```diff
@@ -90,6 +98,15 @@ export class Room extends EventEmitter implements Sayable {
   //   return this.load(contactGetter)
   // }
 
+  private async readyAllMembers(memberList: RoomRawMember[]): Promise<void> {
+    let contact: Contact
```
zixia: get rid of this line

src/room.ts
```diff
@@ -90,6 +98,15 @@ export class Room extends EventEmitter implements Sayable {
...
+    for (let member of memberList) {
+      contact = Contact.load(member.UserName)
```
zixia: change this line to:
`let contact = Contact.load(member.UserName)`
because it's only used in this closure.

src/room.ts
```diff
@@ -114,6 +131,7 @@ export class Room extends EventEmitter implements Sayable {
       const data = await contactGetter(this.id)
       log.silly('Room', `contactGetter(${this.id}) resolved`)
       this.rawObj = data
+      await this.readyAllMembers(this.rawObj.MemberList)
```
zixia: good

src/room.ts
```diff
@@ -209,23 +231,47 @@ export class Room extends EventEmitter implements Sayable {
...
+   */
   private parseNickMap(memberList: RoomRawMember[]): Map<string, string> {
     const nickMap: Map<string, string> = new Map<string, string>()
-
+    let contact: Contact
     if (memberList && memberList.map) {
       memberList.forEach(member => {
-        /**
-         * ISSUE #64 emoji need to be striped
-         * ISSUE #104 never use remark name because sys group message will never use that
```
zixia: Please keep the original comments in the code, and add new comments if needed, for keep the history clear.(until someday we decide to clean them)

src/room.ts
```diff
@@ -209,23 +231,47 @@ export class Room extends EventEmitter implements Sayable {
...
 
+  /**
+   * ISSUE #64 emoji need to be striped
+   * ISSUE #104 never use remark name because sys group message will never use that
+   * @rui: cannot use argument NickName because it mix real nick and remark 
+   */
   private parseNickMap(memberList: RoomRawMember[]): Map<string, string> {
     const nickMap: Map<string, string> = new Map<string, string>()
-
+    let contact: Contact
```
zixia: move `let` to be inside the closure.

src/room.ts
```diff
@@ -209,23 +231,47 @@ export class Room extends EventEmitter implements Sayable {
...
+        nickMap[member.UserName] = UtilLib.stripEmoji(contact.name())
       })
     }
     return nickMap
   }
 
+  private parseRemarkMap(memberList: RoomRawMember[]): Map<string, string> {
+    const remarkMap: Map<string, string> = new Map<string, string>()
+    let contact: Contact
+    let remark: string | null
```
zixia: move `let` to be inside the closure.

src/room.ts
```diff
@@ -209,23 +231,47 @@ export class Room extends EventEmitter implements Sayable {
...
+      memberList.forEach(member => {
+        contact = Contact.load(member.UserName)
+        remark =  contact.remark() || ''
+        remarkMap[member.UserName] = UtilLib.stripEmoji(remark)
+      })
+    }
+    return remarkMap
+  }
+
+  private parseDisplayName(memberList: RoomRawMember[]): Map<string, string> {
```
zixia: parseRemarkMap & parseDisplayName has many duplicated code. 

Please consider to be not duplicated that much.

src/room.ts
```diff
@@ -354,28 +400,49 @@ export class Room extends EventEmitter implements Sayable {
...
   /**
-   * NickName / DisplayName / RemarkName of member
+   * find member by `nick`(NickName) / `display`(DisplayName) / `remark`(RemarkName)
+   * when use member(name:string), find member by nickName by default
    */
-  public member(name: string): Contact | null {
-    log.verbose('Room', 'member(%s)', name)
+  public member(queryArg: MemberQueryFilter | string): Contact | null {
+    let query: MemberQueryFilter
+    if (typeof queryArg === 'string') {
```
zixia: Need to support `member(name: string)` here, which name could be any of remark/display/nick.

src/room.ts
```diff
@@ -354,28 +400,49 @@ export class Room extends EventEmitter implements Sayable {
...
+    const keyMap = {
+      nick:     'nickMap',
+      remark:   'remarkMap',
+      display:  'displayMap'
+    }
 
-    /**
-     * ISSUE #64 emoji need to be striped
-     */
-    name = UtilLib.stripEmoji(name)
```
zixia: Keep the comments.

lijiarui: sorry, I cannot find this position.

src/room.ts
lijiarui: > parseRemarkMap & parseDisplayName has many duplicated code.

> /**
        * ISSUE #64 emoji need to be striped
         * ISSUE #104 never use remark name because sys group message will never use that
         */

Actually, parseRemarkMap & parseDisplayName & parseNickMap has many duplicated code.   So should I not duplicated all?  Then maybe I cannot keep the original comments in the code...

src/puppet-web/firer.ts
lijiarui: I have two questions:

1. room.member() should use which order？  

I have found the rules of the name display order in wechat group's in #173 as follows:

>  Display order:
 @ Event: `displayName > nickName`
 system message `remark > nickName`
 on-screen Names:`remark > displayName > nickName`

I suggest choosing `on-screen Name`, because it contains `system message` and it is useful in `firer.ts`

2. Now room.member can just return one contact, should it return a contact list?

lijiarui:
@zixia  done as what you said.

function `room.member()` use the order : `remark> display>nick`

src/room.ts
```diff
@@ -209,21 +230,40 @@ export class Room extends EventEmitter implements Sayable {
...
 
-  private parseNickMap(memberList: RoomRawMember[]): Map<string, string> {
-    const nickMap: Map<string, string> = new Map<string, string>()
-
+  /**
+   * ISSUE #64 emoji need to be striped
+   * ISSUE #104 never use remark name because sys group message will never use that
+   * @rui: cannot use argument NickName because it mix real nick and remark 
+   */
+  private parseMap(memberList: RoomRawMember[], parseContent: string): Map<string, string> {
```
zixia: Add

`type NameType = 'nick' | 'display' | 'remark'`

Then change 

`parseContent: string`

to

`parseContent: NameType`

src/room.ts
```diff
@@ -209,21 +230,40 @@ export class Room extends EventEmitter implements Sayable {
...
-         * ISSUE #104 never use remark name because sys group message will never use that
-         */
-        nickMap[member.UserName] = UtilLib.stripEmoji(
-          member.DisplayName || member.NickName
-        )
+        let contact = Contact.load(member.UserName)
+        if (parseContent === 'nick') {
+          mapList[member.UserName] = UtilLib.stripEmoji(contact.name())
+        } else if (parseContent === 'remark') {
+          mapList[member.UserName] = UtilLib.stripEmoji(contact.remark() || '')
```
zixia: Change this part of code to this way:

```ts
       let tmpName: string
       switch (parseContent) {
       case 'nick':
          tmpName = contact.name()
          break
       case 'remark':
          tmpName = contact.remark() || ''
          break
       default:
          throw new Error('not found')
       }
       mapList[member.UserName] = UtilLib.stripEmoji(tmpName)
```

src/room.ts
```diff
@@ -209,21 +230,40 @@ export class Room extends EventEmitter implements Sayable {
...
+  private parseNickMap(memberList: RoomRawMember[]): Map<string, string> {
+    return this.parseMap(memberList, 'nick')
+  }
+
+  private parseRemarkMap(memberList: RoomRawMember[]): Map<string, string> {
+    return this.parseMap(memberList, 'remark')
+  }
+
+  private parseDisplayName(memberList: RoomRawMember[]): Map<string, string> {
+    return this.parseMap(memberList, 'display')
```
zixia: Get rid of unnessesary functions: parseNickMap / parseRemarkMap / parseDisplayName

use `this.parseMap(memberList, 'display')` directly

src/room.ts
```diff
@@ -190,6 +207,8 @@ export class Room extends EventEmitter implements Sayable {
 
     const memberList  = this.parseMemberList(rawObj.MemberList)
     const nickMap     = this.parseNickMap(rawObj.MemberList)
+    const remarkMap   = this.parseRemarkMap(rawObj.MemberList)
+    const displayMap  = this.parseDisplayName(rawObj.MemberList)
```
zixia: Instead of introduce new functions, use `this.parseMap(memberList, 'display')` directly at here.

src/room.ts
```diff
@@ -356,31 +396,60 @@ export class Room extends EventEmitter implements Sayable {
...
   /**
-   * NickName / DisplayName / RemarkName of member
+   * find member by `nick`(NickName) / `display`(DisplayName) / `remark`(RemarkName)
+   * when use member(name:string), find member by nickName by default
    */
-  public member(name: string): Contact | null {
-    log.verbose('Room', 'member(%s)', name)
+  public member(queryArg: MemberQueryFilter | string): Contact | null {
+    if (typeof queryArg === 'string') {
+      log.error('Room', 'function member should use member(queryArg: MemberQueryFilter)')
```
zixia: Should not use `log.error()` here because it is not a bug/error, it's a feature.

lijiarui:
@zixia  Thanks for you advice, and I have done what you said.

I'm sorry for not passing  CI tests, I didn't use function `getContact` and I found it is used in function `room.ready()` in the following function:

```
    if (!contactGetter) {
      contactGetter = Config.puppetInstance()
                            .getContact.bind(Config.puppetInstance())
    }
```

Could you help me? Thanks!

src/room.ts
```diff
@@ -356,31 +392,60 @@ export class Room extends EventEmitter implements Sayable {
...
+   * find member by `nick`(NickName) / `display`(DisplayName) / `remark`(RemarkName)
+   * when use member(name:string), find member by nickName by default
    */
-  public member(name: string): Contact | null {
-    log.verbose('Room', 'member(%s)', name)
+  public member(queryArg: MemberQueryFilter | string): Contact | null {
+    if (typeof queryArg === 'string') {
+      log.verbose('Room', 'function member should use member(queryArg: MemberQueryFilter)')
+      return this.member({remark: queryArg}) || this.member({display: queryArg}) || this.member({nick: queryArg})
+    } else {
```
zixia: Remove else here, and get all the code in this block out.

src/room.ts
```diff
@@ -356,31 +392,60 @@ export class Room extends EventEmitter implements Sayable {
   }
 
   /**
-   * NickName / DisplayName / RemarkName of member
+   * find member by `nick`(NickName) / `display`(DisplayName) / `remark`(RemarkName)
+   * when use member(name:string), find member by nickName by default
    */
-  public member(name: string): Contact | null {
-    log.verbose('Room', 'member(%s)', name)
+  public member(queryArg: MemberQueryFilter | string): Contact | null {
```
zixia: Add function declarations before function implement like this:

```diff
+ public member(filter: MemberQueryFilter): Contact | null
+ public member(name: string): Contact | null
+ 
public member(queryArg: MemberQueryFilter | string): Contact | null {
```

src/room.ts
zixia: notice this.

lijiarui:
the following bug is caused by `room.nick(contact: Contact)`, you have defined its return value is `displayName` before, but after I changed the code, its return value is `nickName`.
```
 1. test » room » Room smoking test
  AssertionError: should get nick1 from DisplayName   
    ""              "田美坤"
        test/room.spec.ts:104:5
```

And I changed as you designed before,  function `room.nick()` return displayName, although I think it looks a bit weird。。。。

lijiarui:
it still has an error:

```
✖ test › room › Room smoking test should get nick2 from NickName because there is no DisplayName,    
  t.is(nick2, EXPECTED.memberNick2, 'should get nick2 from NickName because there is no DisplayName, ')
       |               |                                                                               
       ""              "李竹～英诺天使（此号已满）"          
```

Actually, I have tested, using function `room.nick()` if one doesn't have a displayName in the room, it will return nickName correctly. But it failed this test.

I thought maybe I changed some structure of the member obj, so the test data maybe outdated...   
@zixia  need your help, thanks.

zixia:
After you fixed this, I believe this PR could be ready for merging.

lijiarui:
Finally, I found the reason for not passing the test....   

I added `contact.ready()` in new PR, because the contact data in `Class Contact` is not the same as the the member in room, so contact.ready() cannot get the full contact data, then it cannot get the correct nickMap and remarkMap...   

Maybe you should mock some contact data corresponding the room member,  then it can get the full contact data, and get the nickMap, then pass the test...

zixia:
Great. Could you modify the mock data in the unit test to add support for your new code?

test/room.spec.ts
```diff
@@ -55,9 +55,7 @@ Config.puppetInstance(new PuppetWeb())
...
...4739f4","NickName":"vicky张","AttrStatus":100477,"PYInitial":"","PYQuanPin":"","RemarkPYInitial":"","RemarkPYQuanPin":"","MemberStatus":0,"DisplayName":"","KeyWord":"hua","HeadImgUrl":"/cgi-bin/mmwebwx-bin/webwxgeticon?seq=0&username=@ad954bf2159a572b7743a5bc134739f4&skey=@crypt_f9cec94b_f23a307a23231cfb5098faf91ff759ca&chatroomid=@4b8baa99bdfc354443711412126d2aaf"}],"RemarkName":"","HideInputBarFlag":0,"Sex":0,"Signature":"","VerifyFlag":0,"OwnerUin":2351423900,"StarFriend":0,"AppAccountFlag":0,"Statues":0,"AttrStatus":0,"Province":"","City":"","Alias":"","SnsFlag":0,"UniFriend":0,"DisplayName":"","ChatRoomId":0,"KeyWord":"","EncryChatRoomId":"@4b8baa99bdfc354443711412126d2aaf","MMFromBatchGet":true,"MMOrderSymbol":"TONGZHIZHONGGUOQINGNIANTIANSHIHUIGUANFANGJIA","MMFromBatchget":true,"MMInChatroom":true},"@ad85207730aa94e006ddce28f74e6878":{"UserName": "@ad85207730aa94e006ddce28f74e6878","NickName": "田美坤Maggie","RemarkName": ""},"@72c4767ce32db488871fdd1c27173b81":{"UserName":"@72c4767ce32db488871fdd1c27173b81","NickName": "李竹～英诺天使（此号已满）","RemarkName": ""},"@ecff4a7a86f23455dc42317269aa36ab":{"UserName": "@ecff4a7a86f23455dc42317269aa36ab","NickName": "童玮亮","RemarkName": "童玮亮备注"}}`)
```
zixia: Your modification made `RAW_OBJ` data wrong.

RAW_OBJ should be the same structure with the JSON data comes from webwxApp. You can put any data in the struct, as long as your structure is the same as webwxApp.

What you changed here, is just make you happy and make your test code happy. 

Please change the RAW_OBJ with a real room data instead.

test/room.spec.ts
```diff
@@ -111,11 +119,13 @@ test('Room smoking test', async t => {
...
   const contactA = r.member(EXPECTED.memberNick1)
   const contactB = r.member(EXPECTED.memberNick2)
-  if (!contactA || !contactB) {
+  const contactC = r.member(EXPECTED.memberNick3)
+  if (!contactA || !contactB || !contactC) {
     throw new Error('no a or b')
   }
   t.is(contactA.id, EXPECTED.memberId1, 'should get the right id from nick 1')
   t.is(contactB.id, EXPECTED.memberId2, 'should get the right id from nick 2')
+  t.is(contactC.id, EXPECTED.memberId3, 'should get the right id from nick 2')
```
zixia: Please change the descript to more details with your test purpose.

src/room.ts
```diff
@@ -356,26 +390,59 @@ export class Room extends EventEmitter implements Sayable {
...
    */
-  public member(name: string): Contact | null {
-    log.verbose('Room', 'member(%s)', name)
+
+  public member(filter: MemberQueryFilter): Contact | null
+  public member(name: string): Contact | null
+
+  public member(queryArg: MemberQueryFilter | string): Contact | null {
+    if (typeof queryArg === 'string') {
+      log.warn('Room', 'member(%s) DEPRECATED, use member(queryArg: MemberQueryFilter) instead.', queryArg)
```
zixia: Get rid of this log.

`member(string)` should not be DEPRECATED. it is a convenience way to find a contact by match name(display/remark/nick).

src/room.ts
```diff
@@ -356,26 +392,59 @@ export class Room extends EventEmitter implements Sayable {
   }
 
   /**
-   * NickName / DisplayName / RemarkName of member
+   * find member by `nick`(NickName) / `display`(DisplayName) / `remark`(RemarkName)
+   * when use member(name:string), find member by nickName by default
```
zixia: you missed this review.

src/room.ts
```diff
@@ -356,26 +390,59 @@ export class Room extends EventEmitter implements Sayable {
...
+  public member(filter: MemberQueryFilter): Contact | null
+  public member(name: string): Contact | null
+
+  public member(queryArg: MemberQueryFilter | string): Contact | null {
+    if (typeof queryArg === 'string') {
+      log.warn('Room', 'member(%s) DEPRECATED, use member(queryArg: MemberQueryFilter) instead.', queryArg)
+      return this.member({remark: queryArg}) || this.member({display: queryArg}) || this.member({nick: queryArg})
+    }
+
+    log.verbose('Room', 'member({ %s })'
```
zixia: use `log.silly` instead of `verbose` here.

`verbose` should be used only once per function, usually at the beginning of the function, to log the function is run and all parameters it has received.

Any other log should be `silly`

src/room.ts
lijiarui: well, I don't think so, it can confused the user...  especially in the real using case....

but I will do as what you said, it's your code...

lijiarui:
Done as what you said, but I'm a little bit confused by the following code:

```shell
  const mockContactGetter = function (id) {
    return new Promise((resolve, reject) => {
      // if (id !== EXPECTED.id && !(id in CONTACT_LIST)) return resolve({})
      if (id === EXPECTED.id) {
        setTimeout(() => {
          return resolve(RAW_OBJ)
        }, 10)
      }
      if (id in CONTACT_LIST) {
        setTimeout(() => {
          return resolve(CONTACT_LIST[id])
        }, 10)
      }
      return resolve({})
    })
  }
```

it gets the following error:
```shell
✖ Room smoking test Error: Cannot read property 'Symbol(Symbol.iterator)' of undefined
```

But when I change the code to the following:
```shell
  const mockContactGetter = function (id) {
    return new Promise((resolve, reject) => {
      if (id !== EXPECTED.id && !(id in CONTACT_LIST)) return resolve({})
      if (id === EXPECTED.id) {
        setTimeout(() => {
          return resolve(RAW_OBJ)
        }, 10)
      }
      if (id in CONTACT_LIST) {
        setTimeout(() => {
          return resolve(CONTACT_LIST[id])
        }, 10)
      }
    })
  }
```

all works well.

What I change is just adjust  `return resolve({})` order.

I googled the error and find this: [this](http://stackoverflow.com/questions/36295535/angular2-typeerror-cannot-read-property-symbolsymbol-iterator-of-undefined)  
> Initialize items to an empty array.

But I still fell confused..

zixia:
Great work!

It is not perfect, but I think it _mergeable_ now.

BTW:
* You will only feel confusing when you know you do not know.
* You will never feel confusing when you do not know you do not know.

Issue #217, opened 2017-01-24
Prepare to rename the nick/remark/display for contact/room


zixia:
Now we have three name type:
1. `nick`
1. `remark`
1. `display`

When we put the named together, they make us feel confusing, because it is not very clear which one is stands for what.

I think Wechaty should rename those names to the most understand names, instead of the names from the webwxApp code.

What's in my mind now, is that:
1. the name-string set by user his/her-self, should be called `name`(before is `nick`)
1. the name-string set by userself in a room, should be called `nick`(before is `display`). `nick` only belongs to the room.
1. the name-string set by others, should be called `alias` (before is `remark`)

If anyone has any suggestion/recommendation,  please let me know. Thanks.

zixia:
@lijiarui How do you think about this?

lijiarui:
I think it is good, maybe DC can give a good suggestion.

zixia:
According to the reply of Rahim Nathwani from Google:

> In the English version of WeChat, those strings are:
> Name
> My Alias in Group
> Alias
> 
> So I suggest keeping it consistent with that, e.g.
> - bot_name
> - bot_alias
> - user_alias

and

> Perhaps I've misunderstood what the three terms in the GitHub issue mean, but it seems you want to distinguish on two dimensions:
> - name vs. alias
> - self or other (自己或其他人)

I want to keep the names more simple from 3 to 2: get rid of _nick_ , only keep _name_ & _alias_

1. the name-string set by user-self, should be called **name**
1. the name-string set by user-self in a room, should be called **room alias**. **room alias** only belongs to the room.
1. the name-string set by bot for others(original remark), should be called **alias**

If we follow the above new naming design, we will have the following changes for the current code base:

1. `Contact.remark()` will change to `Contact.alias()`
1. `Room.nick()` will change to `Room.alias()`
1. `Room.member()` query key will also need to change to the new name design

zixia:
After a week, I believe that to use `alias` is a good decision. We will not have ~~remark~~ and ~~nick~~ anymore.

PR is welcome.

lijiarui:
I thought it should be designed as follows:

####  `Room.alias()` should return contact's roomAlias
If the contact doesn't set its roomAlias in the room, it should return null.     
Because we can use `contact.name()` to get the contact's name.     
And we can use `contact.alias()` to get the contact's alias. 

#### `Room.member()` query key should consider on two dimensions:

* self-set: {name: string}
* other-set: {alias: string}

[...]

PR #234, opened 2017-02-06
rename the nick/remark/display for contact/room #217

lijiarui:


src/room.ts
```diff
@@ -25,12 +25,12 @@ type RoomObj = {
   topic:      string
   ownerUin:   number
   memberList: Contact[]
-  nickMap:    Map<string, string>
-  remarkMap:  Map<string, string>
-  displayMap: Map<string, string>
+  nameMap:      Map<string, string>
+  aliasMap:     Map<string, string>
+  roomAliasMap: Map<string, string>
```
zixia: Remove `roomAliasMap` here. Because `aliasMap` should be `roomAliasMap`

The room should only in charge of search **alias in room**. Search contact alias should use `Contact.find()` instead.

src/room.ts
```diff
@@ -25,12 +25,12 @@ type RoomObj = {
...
 }
 
-type NameType = 'nick' | 'display' | 'remark'
+type NameType = 'name' | 'alias' | 'roomAlias'
```
zixia: no `roomAlias`. `alias` is for room only.

src/room.ts
```diff
@@ -57,9 +57,9 @@ export type RoomQueryFilter = {
 }
 
 export type MemberQueryFilter = {
-  nick?:    string
-  remark?:  string
-  display?: string
+  name?:      string
+  alias?:     string
+  roomAlias?: string
```
zixia: no roomAlias.

src/room.ts
```diff
@@ -350,11 +350,20 @@ export class Room extends EventEmitter implements Sayable {
     return UtilLib.plainText(this.obj ? this.obj.topic : '')
   }
 
+  // should be deprecated
   public nick(contact: Contact): string {
```
zixia: `nick` should be renamed to `alias`  for the consistency.

src/contact.ts
```diff
@@ -55,8 +55,10 @@ export enum Gender {
 }
 
 export type ContactQueryFilter = {
-  name?: string | RegExp
+  name?:   string | RegExp
+  // remark should be deprecated
   remark?: string | RegExp
```
lijiarui: Then we should deprecate function `Contact.find(remark: string) ` right now?

src/contact.ts
```diff
@@ -232,6 +234,10 @@ export class Contact implements Sayable {
       query = { name: /.*/ }
     }
 
+    if (query[0] === 'remark') {
+      log.warn('Contact', 'findAll(remark:%s) DEPRECATED, use findAll(alias:%s) instead.')
```
lijiarui: no need to add `query[0] = 'alias'`, because I had deal the alias in the following code:
```
const keyMap = {
      name:   'NickName',
      // should be deprecated
      remark: 'RemarkName',
      alias:  'RemarkName',
    }
...
 if (filterValue instanceof RegExp) {
      filterFunction = `(function (c) { return ${filterValue.toString()}.test(c.${filterKey}) })`
    } else if (typeof filterValue === 'string') {
      filterValue = filterValue.replace(/'/g, '\\\'')
      filterFunction = `(function (c) { return c.${filterKey} === '${filterValue}' })`
    } else {
      throw new Error('unsupport name type')
    }
```

If you are sure to deprecate function `Contact.find(remark:string)` right now, I will delete this

src/contact.ts
```diff
@@ -287,46 +295,57 @@ export class Contact implements Sayable {
...
                     return ret
                   })
                   .catch(e => {
-                    log.error('Contact', 'remark(%s) rejected: %s', newRemark, e.message)
+                    log.error('Contact', 'alias(%s) rejected: %s', newAlias, e.message)
                     return false // fail safe
                   })
   }
 
+  public remark(newRemark?: string|null): Promise<boolean> | string | null {
```
lijiarui: should  `log.warn('Contact', 'remark(%s) DEPRECATED, use alias(%s) instead.')` be ok? I have added it under the function

src/contact.ts
zixia: no, you misunderstand.

you should remove `remark` in keyMap first, then you will have to follow my request.

src/contact.ts
zixia: deprecate right now. but it should still functional.

lijiarui: ok, I guess I can use function `Contact.find(query: ContactQueryFilter | {remark: string})` instead, is it right?

src/room.ts
lijiarui: If we use Contact.find() instead, then when use function `member()` to get Contact would be asynchronous， which means it will use 
`public async member(queryArg: MemberQueryFilter | string): Promise<Contact | null>` 
but actually we want to use synchronous function.

And, we must use alias(remark) in room, which I have said in #104 :

You said in [room.ts](https://github.com/wechaty/wechaty/blob/master/src/room.ts)

> ISSUE #104 never use remark name because sys group message will never use that

**But sys group message use remark.** 

For example, when the bot set remark for A, then A changed the topic, then sys message shows the remark, not roomAlias.

I have summarize in [#173](https://github.com/wechaty/wechaty/issues/173):

>

> Display order:
> 
> @ Event
> 
> When someone @ a contact in a room, wechat recognise the name order as follows:
> displayName > nickName
> 
> system message
> 
> When room event happens(join, leave, changetopic), system message recognise the name order as follows:
> remark > nickName
> 
> on-screen Names
> 
> The contact name showed to the wechat user in the group
> remark > displayName > nickName

lijiarui: So，I recommend we should use three kind of map:

> nameMap
> aliasMap
> roomAliasMap

zixia: Ok, you are right. 

Agreed.

src/contact.ts
```diff
@@ -287,51 +292,63 @@ export class Contact implements Sayable {
...
+
   /**
    * try to find a contact by filter: {name: string | RegExp}
    */
-  public static async find(query: ContactQueryFilter): Promise<Contact> {
-    log.verbose('Contact', 'find(%s)', query.name)
+  public static async find(query: ContactQueryFilter | {remark: string | RegExp}): Promise<Contact> {
+    log.verbose('Contact', 'find(%s)', query)
 
     const contactList = await Contact.findAll(query)
```
zixia: should also `log.warn()` for `remark` in `find()` function.

src/room.ts
```diff
@@ -350,11 +351,17 @@ export class Room extends EventEmitter implements Sayable {
...
-    if (!this.obj || !this.obj.nickMap) {
+    log.warn('Room', 'nick(Contact) DEPRECATED, use name(Contact) instead.')
```
zixia: `name(Contact)` here should be `alias(Contact)`?

src/room.ts
```diff
@@ -399,7 +406,7 @@ export class Room extends EventEmitter implements Sayable {
 
   public member(queryArg: MemberQueryFilter | string): Contact | null {
     if (typeof queryArg === 'string') {
-      return this.member({remark: queryArg}) || this.member({display: queryArg}) || this.member({nick: queryArg})
+      return this.member({alias: queryArg}) || this.member({roomAlias: queryArg}) || this.member({name: queryArg})
```
zixia: here `alias` & `roomAlias` is a bit confused.

consider to rename to `contactAlias` & `roomAlias` to declare them more clearly.

src/room.ts
```diff
@@ -423,9 +430,9 @@ export class Room extends EventEmitter implements Sayable {
     let filterValue: string  = UtilLib.stripEmoji(queryArg[filterKey])
 
     const keyMap = {
-      nick:     'nickMap',
-      remark:   'remarkMap',
-      display:  'displayMap'
+      name:       'nameMap',
+      alias:      'aliasMap',
```
zixia: also use `contactAlias` & `roomAlias` at here.

lijiarui:
I guess Coverage decrease because I deleted `const name2 = r.name(contact2)` in `room.spec.ts`, because it is a wrong function caused by me in the first commit.

I'm sorry in my first commit I add a wrong function `Room.name(contact)`, but after read your request, I changed function `name()` into `alias()`.

src/contact.ts
lijiarui: Well, I didn't add log warn  because `find()` use `findAll()`, and I had add `log.warn()` for  `findAll()`. 

Do you agree?

src/room.ts
```diff
@@ -353,15 +347,15 @@ export class Room extends EventEmitter implements Sayable {
...
   public nick(contact: Contact): string {
-    log.warn('Room', 'nick(Contact) DEPRECATED, use name(Contact) instead.')
+    log.warn('Room', 'nick(Contact) DEPRECATED, use alias(Contact) instead.')
     return this.alias(contact)
   }
 
   public alias(contact: Contact): string {
-    if (!this.obj || !this.obj.nameMap) {
+    if (!this.obj) {
       return ''
```
zixia: should `return null` be better?

src/room.ts
```diff
@@ -353,15 +347,15 @@ export class Room extends EventEmitter implements Sayable {
...
     }
-    return this.obj.roomAliasMap[contact.id] || this.obj.nameMap[contact.id]
+    return this.obj.aliasMap[contact.id]
```
zixia: if no `aliasMap[contact.id]`, here'll return a `undefined`, which is not expected?

src/contact.ts
```diff
@@ -222,12 +222,17 @@ export class Contact implements Sayable {
...
+   * find contact by `name` or `alias`
    */
-  public static async findAll(queryArg?: ContactQueryFilter): Promise<Contact[]> {
+  public static async findAll(queryArg?: ContactQueryFilter | {remark: string | RegExp}): Promise<Contact[]> {
     let query: ContactQueryFilter
     if (queryArg) {
-      query = queryArg
+      if (Object.keys(queryArg)[0] === 'remark') {
+        log.warn('Contact', 'Contact.findAll(remark:%s) DEPRECATED, use Contact.findAll(alias:%s) instead.')
+        query = { alias: queryArg['remark']}
```
zixia: queryArg.remark

src/contact.ts
```diff
@@ -222,12 +222,17 @@ export class Contact implements Sayable {
...
   /**
-   * find contact by `name`(NickName) or `remark`(RemarkName)
+   * find contact by `name` or `alias`
    */
-  public static async findAll(queryArg?: ContactQueryFilter): Promise<Contact[]> {
+  public static async findAll(queryArg?: ContactQueryFilter | {remark: string | RegExp}): Promise<Contact[]> {
     let query: ContactQueryFilter
     if (queryArg) {
-      query = queryArg
+      if (Object.keys(queryArg)[0] === 'remark') {
```
zixia: `if (queryArg.remark) {`

src/contact.ts
```diff
@@ -287,51 +292,63 @@ export class Contact implements Sayable {
...
 
+  // function should be deprecated
+  public remark(newRemark?: string|null): Promise<boolean> | string | null {
+    log.warn('Contact', 'remark(%s) DEPRECATED, use alias(%s) instead.')
+    log.silly('Contact', 'remark(%s)', newRemark || '')
+
+    if (newRemark) {
+      return this.alias(newRemark)
+    } else {
+      return this.alias()
```
zixia: when `newRemark = null`, will run code here, which is not expected.

src/room.ts
```diff
@@ -206,10 +204,9 @@ export class Room extends EventEmitter implements Sayable {
       return null
     }
 
-    const memberList  = this.parseMemberList(rawObj.MemberList)
-    const nickMap     = this.parseMap(rawObj.MemberList, 'nick')
-    const remarkMap   = this.parseMap(rawObj.MemberList, 'remark')
-    const displayMap  = this.parseMap(rawObj.MemberList, 'display')
+    const memberList = this.parseMemberList(rawObj.MemberList)
+    const nameMap    = this.parseMap(rawObj.MemberList, 'nick')
```
zixia: why `nameMap` use arg name with `nick` here?

should be consistent.

src/room.ts
```diff
@@ -238,21 +234,20 @@ export class Room extends EventEmitter implements Sayable {
         let contact = Contact.load(member.UserName)
         switch (parseContent) {
           case 'nick':
```
zixia: confusing names & logic here...

src/room.ts
```diff
@@ -25,12 +25,11 @@ type RoomObj = {
...
   memberList: Contact[]
-  nameMap:      Map<string, string>
-  aliasMap:     Map<string, string>
-  roomAliasMap: Map<string, string>
+  nameMap:    Map<string, string>
+  aliasMap:   Map<string, string>
 }
 
-type NameType = 'name' | 'alias' | 'roomAlias'
+type NameType = 'nick' | 'alias'
```
lijiarui: sorry, my fault..

src/contact.ts
```diff
@@ -227,9 +227,9 @@ export class Contact implements Sayable {
   public static async findAll(queryArg?: ContactQueryFilter | {remark: string | RegExp}): Promise<Contact[]> {
     let query: ContactQueryFilter
     if (queryArg) {
-      if (queryArg[0] === 'remark') {
+      if (Object.keys(queryArg)[0] === 'remark') {
```
lijiarui: I tried it before, but it  isn't compiled successfully:
```
 public static async findAll(queryArg?: ContactQueryFilter | {remark: string | RegExp}): Promise<Contact[]> {
    let query: ContactQueryFilter
    if (queryArg) {
      if (queryArg.remark) {
        query = { alias: queryArg.remark}
      }
```
 error throw:
```
src/contact.ts(230,20): error TS2339: Property 'remark' does not exist on type 'ContactQueryFilter | { remark: string | RegExp; }'.
  Property 'remark' does not exist on type 'ContactQueryFilter'.
src/contact.ts(231,35): error TS2339: Property 'remark' does not exist on type 'ContactQueryFilter | { remark: string | RegExp; }'.
  Property 'remark' does not exist on type 'ContactQueryFilter'.
```

src/contact.ts
```diff
@@ -227,9 +227,9 @@ export class Contact implements Sayable {
...
         log.warn('Contact', 'Contact.findAll(remark:%s) DEPRECATED, use Contact.findAll(alias:%s) instead.')
-        query = { alias: queryArg[1]}
+        query = { alias: queryArg['remark']}
```
lijiarui: it will throw error

src/room.ts
lijiarui: if no `aliasMap[contact.id]` here, it should return undefiend. Actually, I have deal alias in aliasMap as follows:
```ts
private parseMap(memberList: RoomRawMember[], parseContent: NameType): Map<string, string> {
    const mapList: Map<string, string> = new Map<string, string>()
    if (memberList && memberList.map) {
      memberList.forEach(member => {
        let tmpName: string
        let contact = Contact.load(member.UserName)
        switch (parseContent) {
          case 'name':
            tmpName = contact.alias() || contact.name()
            break
          case 'alias':
            tmpName = member.DisplayName || contact.name()
            break
          default:
            throw new Error('parseMap failed, member not found')
        }
```

if no `DisplayName`, this code `tmpName = member.DisplayName || contact.name()` will get `contact.name()`   

So, no need to add `this.obj.nameMap[contact.id]` here, what do you think about it?

src/contact.ts
zixia: ok

src/contact.ts
zixia: ok

src/room.ts
```diff
@@ -350,11 +345,17 @@ export class Room extends EventEmitter implements Sayable {
...
+    log.warn('Room', 'nick(Contact) DEPRECATED, use alias(Contact) instead.')
+    return this.alias(contact)
+  }
+
+  public alias(contact: Contact): string | null {
+    if (!this.obj) {
+      return null
     }
-    return this.obj.displayMap[contact.id] || this.obj.nickMap[contact.id]
+    return this.obj.aliasMap[contact.id]
```
zixia: return this.obj.aliasMap[contact.id] || null

src/contact.ts
```diff
@@ -222,12 +222,17 @@ export class Contact implements Sayable {
   }
 
   /**
-   * find contact by `name`(NickName) or `remark`(RemarkName)
+   * find contact by `name` or `alias`
    */
-  public static async findAll(queryArg?: ContactQueryFilter): Promise<Contact[]> {
+  public static async findAll(queryArg?: ContactQueryFilter | {remark: string | RegExp}): Promise<Contact[]> {
```
zixia: move `remark: string | RegExp` to `ContactQueryFilter` with a comment noticing _DEPRECATED_

then you could use `.remark` again

lijiarui: > there's no reason to keep remark here anymore. remove it.

you just told me to remove remark from ContactQueryFilter

zixia: Ok, let me tell u one more:

Next time, if u still only know what instead of knowing why, and make stupid argue like this, then kick your ass 10 times by yourself.

Issue #251, opened 2017-02-14
[linting] fix needed for new tslint rule: trailing-comma


zixia:
```shell
src/util-lib.ts[221, 26]: Missing trailing comma
```

lijiarui:
@zixia please close this issue

zixia:
You can close an issue by put `close #251` in your commit message.

Once your commit is merged, the referred issue will be closed automatically.

lijiarui:
Thanks! I will try it next time! 

Could you please to help close the issue this time? because I just delete the branch I commited...

zixia:
It's already closed.

PR #261, opened 2017-02-16
Add friendrequest sys message

lijiarui:
#260


zixia:
Hi,

Before pushing any new PR should fix #251 first because otherwise, the TsLint will always fail.

Thanks.

Issue #273, opened 2017-02-22
Cannot read property 'Symbol(Symbol.iterator)' of undefined


Whuhxh:
1.Run npm run doctor or wechaty run doctor(for docker user), paste output here

node -v: 6.9.1
npm -v: 3.10.8

  #### Wechaty Doctor

  1. Wechaty version: #git[7faecd6 0.7.32]
  2. uname -r: 3.10.0-327.10.1.el7.x86_64
  3. Docker: false
  4. Node version: v6.9.1

2.Expected behavior

When start bot with calling Contact.load(id) to load a contact, but the contact id is some "undefined" value, the bot should deal with id of "undefined" situation.

3.Actual behavior

ERR Room contactGetter(@@c4f424080c5f480442dfb0dd5af062b70443edf95c9d1b87c1e15ab1b212fe24) exception: Cannot read property 'Symbol(Symbol.iterator)' of undefined
ERR Message ready() exception: TypeError: Cannot read property 'Symbol(Symbol.iterator)' of undefined
at Room. (/bot/node_modules/wechaty/dist/src/room.js:52:32)
at Generator.next ()
at /bot/node_modules/wechaty/dist/src/room.js:7:71
at __awaiter (/bot/node_modules/wechaty/dist/src/room.js:3:12)
at Room.readyAllMembers (/bot/node_modules/wechaty/dist/src/room.js:51:16)
at Room. (/bot/node_modules/wechaty/dist/src/room.js:88:28)
at Generator.next ()
at fulfilled (/bot/node_modules/wechaty/dist/src/room.js:4:58)
at process._tickDomainCallback (internal/process/next_tick.js:129:7)
ERR PuppetWebEvent onServerMessage() exception: TypeError: Cannot read property 'Symbol(Symbol.iterator)' of undefined
at Room. (/bot/node_modules/wechaty/dist/src/room.js:52:32)
at Generator.next ()
at /bot/node_modules/wechaty/dist/src/room.js:7:71
at __awaiter (/bot/node_modules/wechaty/dist/src/room.js:3:12)
at Room.readyAllMembers (/bot/node_modules/wechaty/dist/src/room.js:51:16)
at Room. (/bot/node_modules/wechaty/dist/src/room.js:88:28)
at Generator.next ()
at fulfilled (/bot/node_modules/wechaty/dist/src/room.js:4:58)
at process._tickDomainCallback (internal/process/next_tick.js:129:7)
ERR Config ###########################
[...]

PR #275, opened 2017-02-22
load all memberList

lijiarui:
#273

### DIFF
--- a/src/room.ts
+++ b/src/room.ts
@@ -14,9 +14,9 @@ type RoomObj = {
   encryId:    string,
   topic:      string,
   ownerUin:   number,
-  memberList?: Contact[],
-  nameMap?:    Map<string, string>,
-  aliasMap?:   Map<string, string>,
+  memberList: Contact[],
+  nameMap:    Map<string, string>,
+  aliasMap:   Map<string, string>,
 }
 
 type NameType = 'name' | 'alias'
@@ -131,15 +131,12 @@ export class Room extends EventEmitter implements Sayable {
       log.silly('Room', `contactGetter(${this.id}) resolved`)
       this.rawObj = data
       if (!this.rawObj.MemberList) {
-        this.rawObj.MemberList = []
+        throw new Error('no this.obj.memberList, set after contactGetter')
       }
       await this.readyAllMembers(this.rawObj.MemberList)
       this.obj    = this.parse(this.rawObj)
       if (!this.obj) {
         throw new Error('no this.obj set after contactGetter')
-      }
-      if (!this.obj.memberList) {
-        this.obj.memberList = []
       }
       await Promise.all(this.obj.memberList.map(c => c.ready(contactGetter)))
 
@@ -209,15 +206,8 @@ export class Room extends EventEmitter implements Sayable {
       log.warn('Room', 'parse() on a empty rawObj?')
       return null
     }
-
     if (!rawObj.MemberList) {
-      return {
-        id:         rawObj.UserName,
-        encryId:    rawObj.EncryChatRoomId, // ???
-        topic:      rawObj.NickName,
-        ownerUin:   rawObj.OwnerUin,
-        memberList: [],
-      }
+      rawObj.MemberList = []
     }
 
     const memberList = this.parseMemberList(rawObj.MemberList)
--- a/src/room.ts
+++ b/src/room.ts
@@ -131,7 +131,7 @@ export class Room extends EventEmitter implements Sayable {
       log.silly('Room', `contactGetter(${this.id}) resolved`)
       this.rawObj = data
       if (!this.rawObj.MemberList) {
-         return this.ready()
+        this.rawObj.MemberList = []
       }
       await this.readyAllMembers(this.rawObj.MemberList)
       this.obj    = this.parse(this.rawObj)
@@ -139,7 +139,7 @@ export class Room extends EventEmitter implements Sayable {
         throw new Error('no this.obj set after contactGetter')
       }
       if (!this.obj.memberList) {
-        throw new Error('no this.obj.memberLit set after contactGetter')
+        this.obj.memberList = []
       }
       await Promise.all(this.obj.memberList.map(c => c.ready(contactGetter)))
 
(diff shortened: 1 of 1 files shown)


lijiarui:
Then, what's your suggestion?

zixia:
My suggestion is: keep thinking hard, instead of keep asking.

zixia:
The best practice of making Pull Request is to keep improving it instead of closing it because close it will lose all the review information.

src/room.ts
```diff
@@ -14,9 +14,9 @@ type RoomObj = {
   encryId:    string,
   topic:      string,
   ownerUin:   number,
-  memberList: Contact[],
-  nameMap:    Map<string, string>,
-  aliasMap:   Map<string, string>,
+  memberList?: Contact[],
+  nameMap?:    Map<string, string>,
```
zixia: Right.

src/room.ts
```diff
@@ -130,11 +130,17 @@ export class Room extends EventEmitter implements Sayable {
       const data = await contactGetter(this.id)
       log.silly('Room', `contactGetter(${this.id}) resolved`)
       this.rawObj = data
+      if (!this.rawObj.MemberList) {
+         return this.ready()
```
zixia: You should not do this because this will cause very bad problems.

src/room.ts
```diff
@@ -130,11 +130,17 @@ export class Room extends EventEmitter implements Sayable {
...
+      }
       await this.readyAllMembers(this.rawObj.MemberList)
       this.obj    = this.parse(this.rawObj)
       if (!this.obj) {
         throw new Error('no this.obj set after contactGetter')
       }
+      if (!this.obj.memberList) {
+        throw new Error('no this.obj.memberLit set after contactGetter')
```
zixia: Should not throw here.

memberList is empty should be treated a normal situation, because the data is **async**.

If this.rawObj.MemberList is `undefined`, we should init `obj.memberList` with `[]`.

src/room.ts
```diff
@@ -204,19 +210,27 @@ export class Room extends EventEmitter implements Sayable {
...
+      }
+    }
 
     return {
       id:         rawObj.UserName,
       encryId:    rawObj.EncryChatRoomId, // ???
       topic:      rawObj.NickName,
       ownerUin:   rawObj.OwnerUin,
-
-      memberList,
```
zixia: If this.rawObj.MemberList is `undefined`, we should init **memberList** with `[]`.

src/room.ts
lijiarui: Oh, sorry, I added before and forgot to delete it.......

src/room.ts
zixia: Why I said "right" is because you starting to notice the right problem.

But what you did modify here is not correct.

the `memberList`, `nameMap` and `aliasMap` should not defined by `?`, which means they should always not to be `undefined` or `null`.

If there's no data for them, they should be empty array, or a empty `{}`.

src/room.ts
```diff
@@ -130,11 +130,17 @@ export class Room extends EventEmitter implements Sayable {
       const data = await contactGetter(this.id)
       log.silly('Room', `contactGetter(${this.id}) resolved`)
       this.rawObj = data
+      if (!this.rawObj.MemberList) {
+        this.rawObj.MemberList = []
```
zixia: Do not modify `rawObj`, because this data comes from Browser, which should be treated as _immutable_. 

Do this job on `this.obj` instead.

src/room.ts
```diff
@@ -130,11 +130,17 @@ export class Room extends EventEmitter implements Sayable {
...
+      }
       await this.readyAllMembers(this.rawObj.MemberList)
       this.obj    = this.parse(this.rawObj)
       if (!this.obj) {
         throw new Error('no this.obj set after contactGetter')
       }
+      if (!this.obj.memberList) {
+        this.obj.memberList = []
```
zixia: Do not do this job here.

Do this job inside `this.parse` instead.

src/room.ts
```diff
@@ -204,6 +210,16 @@ export class Room extends EventEmitter implements Sayable {
       return null
     }
 
+    if (!rawObj.MemberList) {
```
zixia: Get rid of this `if`, and also get rid of `return` inside it.

We should prevent to use multiple return statement for the final data.

Only return the final data at the end of function statements. The "return"s  before that should only return for the error conditions.

src/room.ts
```diff
@@ -130,6 +130,9 @@ export class Room extends EventEmitter implements Sayable {
       const data = await contactGetter(this.id)
       log.silly('Room', `contactGetter(${this.id}) resolved`)
       this.rawObj = data
+      if (!this.rawObj.MemberList) {
+        throw new Error('no this.obj.memberList, set after contactGetter')
```
zixia: As I said before: should NOT throw when `!this.rawObj.MemberList`.

Because this is OK.

src/room.ts
```diff
@@ -130,6 +130,9 @@ export class Room extends EventEmitter implements Sayable {
...
+      }
       await this.readyAllMembers(this.rawObj.MemberList)
```
zixia: await this.readyAllMembers(this.rawObj.MemberList || [])

src/room.ts
```diff
@@ -203,6 +206,9 @@ export class Room extends EventEmitter implements Sayable {
       log.warn('Room', 'parse() on a empty rawObj?')
       return null
     }
+    if (!rawObj.MemberList) {
+      rawObj.MemberList = []
```
zixia: Delete them.

As I mentioned before, you should NOT change any value of `rawObj`.

src/room.ts
```diff
@@ -130,7 +130,7 @@ export class Room extends EventEmitter implements Sayable {
       const data = await contactGetter(this.id)
       log.silly('Room', `contactGetter(${this.id}) resolved`)
       this.rawObj = data
-      await this.readyAllMembers(this.rawObj.MemberList)
+      await this.readyAllMembers(this.rawObj.MemberList || [])
```
zixia: Do not do it here.

Do it inside `readyAllMembers()`

src/room.ts
```diff
@@ -204,19 +204,18 @@ export class Room extends EventEmitter implements Sayable {
       return null
     }
 
-    const memberList = this.parseMemberList(rawObj.MemberList)
-    const nameMap    = this.parseMap(rawObj.MemberList, 'name')
-    const aliasMap   = this.parseMap(rawObj.MemberList, 'alias')
+    const memberList = this.parseMemberList(rawObj.MemberList || [])
+    const nameMap    = this.parseMap(rawObj.MemberList || [], 'name')
+    const aliasMap   = this.parseMap(rawObj.MemberList || [], 'alias')
```
zixia: The above three: Do not do it here.

Do it inside `parse*()` function.

src/room.ts
```diff
@@ -204,19 +204,18 @@ export class Room extends EventEmitter implements Sayable {
...
       id:         rawObj.UserName,
       encryId:    rawObj.EncryChatRoomId, // ???
       topic:      rawObj.NickName,
       ownerUin:   rawObj.OwnerUin,
-
-      memberList,
-      nameMap,
-      aliasMap,
+      memberList: memberList,
+      nameMap:  nameMap,
```
zixia: Should keep using `memberList` instead of `memberList: memberList`.

PR #280, opened 2017-02-24
Add print nodejs version

xinbenlv:


zixia:
Thanks!

However this PR will only update the docker image after @lijiarui fixed #273 due to a typescript data structure change, which caused currently the CI isalways fail.

Issue #283, opened 2017-02-25
Room.alias() should return null if we have not set the alias in the room 


zixia:
Now it will return self name if no alias found . It's not right and need to be fixed.

lijiarui:
after you merged #275 I will fix this one.

zixia:
Thanks @lijiarui for fixing this.

PR #303, opened 2017-03-07
#283

lijiarui:
let room.alias() return null when not set roomAlias


src/room.ts
```diff
@@ -222,7 +222,7 @@ export class Room extends EventEmitter implements Sayable {
             tmpName = contact.alias() || contact.name()
             break
           case 'alias':
-            tmpName = member.DisplayName || contact.name()
+            tmpName = member.DisplayName
```
zixia: Should update JSDoc at https://github.com/wechaty/wechaty/pull/303/files#diff-a260093009080ad4951308e3d5ea54fdR342

zixia:
Thanks.

lijiarui:
fix #283

PR #323, opened 2017-03-13
remove m.send() fucntion

lijiarui:


example/friend-bot.ts
```diff
@@ -62,13 +62,13 @@ bot
  */
 .on('friend', async (contact, request) => {
   let logMsg
-  const m = new Message()
-  m.set('to', 'filehelper')
+  const fileTransfer = await Contact.find({name: '文件传输助手'}) || await Contact.find({name: 'File Transfer'})
```
zixia: What if I'm a Traditional Chinese user?
Your code will not work anymore.

What if there's a normal user named "文件传输助手"?
Your code will not work as you expected.

I believe the best way is to find by id.

zixia:
@lijiarui Please leave a comment for noting the reason before close a PR/ISSUE.

zixia:
I'm a little bit disappointed for your giving up so easy, without trying harder.

You could do the very similar thing as `msg.to('filehelper')` by `Contact.load('filehelper')`.

lijiarui:
I

example/friend-bot.ts
```diff
@@ -62,13 +62,13 @@ bot
  */
 .on('friend', async (contact, request) => {
   let logMsg
-  const m = new Message()
-  m.set('to', 'filehelper')
+  const fileTransfer = Contact.load('filehelper')
```
zixia: fileHelper is better

example/friend-bot.ts
```diff
@@ -62,13 +62,13 @@ bot
...
 
   try {
     logMsg = 'received `friend` event from ' + contact.get('name')
-    m.set('content', logMsg)
-    bot.send(m)
+    if (fileTransfer) {
```
zixia: `if` is not necessary here.

Because the `Contact.load()` will always return the contact, or it will throw exception.

example/friend-bot.ts
```diff
@@ -99,8 +99,9 @@ bot
   }
 
   console.log(logMsg)
-  m.set('content', logMsg)
-  bot.send(m)
+  if (fileTransfer) {
```
zixia: get rid of `if`

example/friend-bot.ts
lijiarui: no, I just find in my wechat, it says:
```
WeChat ID: filehelper
```

zixia:
it's so easy ha?

i hope you could learn from this that a good collaborator should think more, talk less and do the exact what you need to do.

lijiarui:
Thanks

PR #358, opened 2017-03-24
first item of memberList as owner is confusion

JasLin:
1. in wxwebapp,first item in memberList is a random user from the room ,it's confusion, 
2. some times we can got owneruin or ChatRoomOwner .

 as those reason, first member from memberlist should not be  here any more


zixia:
Thanks bro, you are right.

PR #287, opened 2017-02-28
add m.mention()

lijiarui:
as #216 told：

add
```
message.mention(): boolean
```
And now it doesn't work if the bot change it's roomAlias in a room.

But I'm not sure what to do when the bot change it's roomAlias in a room...

This remind me if we want to set a permanent id for a room, we suggest not set the bot's alias an `uuid` or md5, or if someone @bot, it will seem something like this `@7465694230576`, really strange.

PR #362, opened 2017-03-26
Add mention

lijiarui:
continue with #287 

I'm trying to reset the merge file. Becasue the `ding` error is not casued by this commit, so I should reset the commit  3c9e1ea `Merge branch 'master' of https://github.com/wechaty/wechaty into add_mention`

As I have said in #287, thanks god, finally I mocked room data successfully...


lijiarui:
I feel a little bit strange for the continuous error:
```shell
src » message » ready() contact ready for ToUserName
```
I feel strange for two reasons:
- it tests well on my computer, but it failed in the test.
- I didn't change any function about   `m.ready()` but it occurs test failed...

src/message.ts
```diff
@@ -403,6 +403,37 @@ export class Message implements Sayable {
     return fromId === userId
   }
 
+  /**
+   * Get the mentioned contact which the message is mentioned for.
+   * @returns {Contact[]} return the contactList which the message is mentioned for
```
zixia: add a `@example` here.

src/message.ts
```diff
@@ -403,6 +403,37 @@ export class Message implements Sayable {
...
+   */
+  public mention(): Contact[] {
+    let contactList: Contact[] = []
+    const room = this.room()
+    if (!room) return contactList
+    if (this.type() !== MsgType.TEXT) return contactList
```
zixia: Please always use `{}` for `if`, even if there's only one line of code, because that will make code more readable.

```ts
if (this.type() !== MsgType.TEXT
    || !this.room) {
    return contactList
}
```

src/message.ts
```diff
@@ -403,6 +403,37 @@ export class Message implements Sayable {
...
+
+    const foundList = this.content().match(/@\S+ ?/g)
+    if (!foundList) return contactList
+
+    const mentionList = foundList.map(element => {
+      /**
+       * fake '@' return element.slice(1, -1), wechat real '@' return element.slice(1)
```
zixia: Can not understand your comment.

Please provide a example in comments to make it clear.

src/message.ts
```diff
@@ -403,6 +403,37 @@ export class Message implements Sayable {
...
+       * fake '@': @ event is produced by typeing '@lijiarui '
+       * real '@': @ event is produced by long press the contact's avatar
+       */
+      return element.slice(1)
+    })
+    log.verbose('Message', 'mention(%s),get mentionList: %s', this.content(), JSON.stringify(mentionList))
+    mentionList.forEach(name => {
+      const contact = room.member({alias: name}) || room.member({name: name})
```
zixia: I believe here should use `room.member(name)` instead of your code, which will require you to fix the `member()` method first.

src/message.ts
```diff
@@ -403,6 +403,37 @@ export class Message implements Sayable {
...
+   * Get the mentioned contact which the message is mentioned for.
+   * @returns {Contact[]} return the contactList which the message is mentioned for
+   */
+  public mention(): Contact[] {
+    let contactList: Contact[] = []
+    const room = this.room()
+    if (!room) return contactList
+    if (this.type() !== MsgType.TEXT) return contactList
+
+    const foundList = this.content().match(/@\S+ ?/g)
```
zixia: Here, the foundList will contains all the strings following by a `@` character?

I suggest renaming `foundList` to `atStringList` to make it more clear and readable.

src/message.ts
```diff
@@ -403,6 +403,37 @@ export class Message implements Sayable {
...
+       * real '@': @ event is produced by long press the contact's avatar
+       */
+      return element.slice(1)
+    })
+    log.verbose('Message', 'mention(%s),get mentionList: %s', this.content(), JSON.stringify(mentionList))
+    mentionList.forEach(name => {
+      const contact = room.member({alias: name}) || room.member({name: name})
+      if (contact) {
+        contactList.push(contact)
+      }
```
zixia: We need to make sure the code is running as expected.

```ts
    if (contact) {
      contactList.push(contact)
    } else {
      log.warn('Message', 'mention() can not found room.member() from mentionList')
      // this will help us to track the unexpected strings.
    }
```

src/message.spec.ts
```diff
@@ -127,3 +127,125 @@ test('self()', t => {
...
+    puppet1.getContact = mockContactGetter
+  } catch (err) {
+    puppet1 = { getContact: mockContactGetter }
+    Config.puppetInstance(puppet1)
+  }
+  const msg11 = new Message(rawObj11)
+  const room11 = msg11.room()
+  if (room11) {
+    await room11.ready()
+    setTimeout(function () {
```
zixia: You'd better find out why you need `setTimeout` at here, or you will be more likely to be bitten again and again.

src/message.ts
lijiarui: cannot use `this.room` instead of `room` here, or typescript will see `this.room` is null.

src/message.ts
lijiarui: yes, the foundeList will contains all the **shortest** string following by a `@` character.

If ones' wechat name is `lijiarui@orange`, it will return `orange`, instead of `lijiarui@orange`.

Or, we should permutation and combination all the possibility.....

What do you think about it?

src/message.ts
lijiarui: here should never use `room.member(name)`, because `room.member(name)` will get all three kinds of names here. But we just use two of them, and if parse `room.member({contactAlias: name})` will get the contact which we doesn't meant to parse.

src/message.ts
```diff
@@ -414,6 +414,59 @@ export class Message implements Sayable {
...
+   * ```
+   */
+  public mention(): Contact[] {
+    let contactList: Contact[] = []
+    const room = this.room()
+    if (this.type() !== MsgType.TEXT || !room ) {
+      return contactList
+    }
+
+    const atStringList = this.content().match(/@\S+ ?/g)
```
xinbenlv: have you considered add the type enforcement of atStringList? If so, why not?

const atStringList:Array<String> = this.content().match(...)

src/message.ts
```diff
@@ -414,6 +414,59 @@ export class Message implements Sayable {
...
+      /**
+       * `fake@` and `real@` definition
+       * Supposed a wechat contact called `lijiarui`
+       * `fake@` means @ event is produced by typing '@lijiarui ' (mock mention behaviour by web wechat)
+       * `real@` means @ event is produced by long press the contact's avatar (the real behaviour in cellphone)
+       *
+       * `fake@` return element.slice(1, -1), `real@` return element.slice(1)
+       */
+      return element.slice(1)
+    })
```
xinbenlv: ;
https://github.com/basarat/typescript-book/blob/master/docs/styleguide/styleguide.md#semicolons

src/message.ts
```diff
@@ -414,6 +414,59 @@ export class Message implements Sayable {
...
+    if (!atStringList) return contactList
+
+    const mentionList = atStringList.map(element => {
+      /**
+       * `fake@` and `real@` definition
+       * Supposed a wechat contact called `lijiarui`
+       * `fake@` means @ event is produced by typing '@lijiarui ' (mock mention behaviour by web wechat)
+       * `real@` means @ event is produced by long press the contact's avatar (the real behaviour in cellphone)
+       *
+       * `fake@` return element.slice(1, -1), `real@` return element.slice(1)
```
xinbenlv: This explains how fake@ works but how do you plan to use it?
Are we saying we will only handle the real@ case?

src/message.ts
```diff
@@ -414,6 +414,59 @@ export class Message implements Sayable {
...
+      if (atRoomAliasList || atContactAliasList) {
+        if (atRoomAliasList) {
+          contactList.concat(atRoomAliasList)
+        }
+        if (atContactAliasList) {
+          contactList.concat(atContactAliasList)
+        }
+      } else {
+        log.warn('Message', 'mention() can not found room.member() from mentionList')
+        // this will help us to track the unexpected strings.
```
xinbenlv: If we want to use them to track the unexpected strings, I'd suggest we also add name here
 log.warn('Message', `mention() can not found room.member() from mentionList given mention name being ${name}`)

Using \`\` you can do string literal inclusion like 
```
`Contact ${name} is being mentioned`;
```

src/message.ts
lijiarui: fake@ doesn't work, it just seems @happens but actually, it won't really mention the user. So I don't plan to use it.
and yes, here we just handle the real@ case.

src/message.ts
lijiarui: Do you mean `return element.slice(1)` here?

src/message.ts
lijiarui: ok! thanks, I'll modify here.

src/message.ts
lijiarui: besides a name with space, we also cannot handle a contact name with `@`, and I suggest to wait until next PR if it really occurs so frequently.
Do you have some good suggestion?

lijiarui: ok, I'll modify it here.

src/message.ts
```diff
@@ -414,6 +414,59 @@ export class Message implements Sayable {
...
+   *
+   * @example
+   * ```ts
+   * const content = message.content()
+   * const contactList = message.mention()
+   * console.log(content)
+   * console.log(contactList)
+   * ```
+   */
+  public mention(): Contact[] {
```
Gcaufy: This may be an English issue, First I see the method name, I thought it would be `mention(somebody, something)`. but after I read the comments, I realize it's getting a mentioned list. 
I know `mention` can be a noun, but still make me confused. If it doesn't make you guys confused, then ignore this comment.

lijiarui: I think you are right, do you have some suggestion?
how about `getMention()` or `isMention()` or `mentioned()`
@dcsan about the function name, do you have any suggestion?

dcsan: `getMentioned()` if that's what it does

`mention()` sounds like it would send an @mention message to the room like 

`mention(user)` sending `@username`

lijiarui: yep, your are right, but `getMentioned()` seems too long. 
@zixia @Gcaufy how do you think about the name?

zixia: I vote for `mention()` because it's also a noun: _n. 提及；说起_

src/message.ts
lijiarui: @zixia maybe you should explain why the entire project doesn't use the semicolon...

src/message.ts
xinbenlv: Since your input is only the message.content(), how do you distinguish the fake-at case and the real-at case? If I understand correctly, by sending a fake-at message with hand typed [at], the current implementation will still trigger the `on('mention')` event. And this function will returns a wrong user-name, right?

src/message.ts
xinbenlv: Right, that's what I am asking. 
We understand that for backward compatibility, javascript supports omitting the semicolon.

src/message.ts
xinbenlv: Agree, then I personally suggest putting a TODO /  FIXIT for future review

src/message.ts
Gcaufy: I vote for `mentioned()`
`mention` can be a noun, but most time it remind me `mention(somebody)`
wechaty api are simple, like `user.self()` instead of `user.isSelf()`, so `getMentioned` is not wechaty style.
but I think `metioned` is fine, it's a adj, can be who are metioned is this room. will not confused me.

lijiarui: it seems `mentioned()` look nice, I vote for this too.

zixia: Agree for using mentioned.

It's more like a wechaty style name than others.

src/message.ts
zixia: @xinbenlv It's just my personal favorite.

And I'm also supporting to not use semicolons, because of:
1. <http://blog.izs.me/post/2353458699/an-open-letter-to-javascript-leaders-regarding>
1. <http://inimino.org/~inimino/blog/javascript_semicolons>
1. <https://www.youtube.com/watch?v=gsfbh17Ax9I>

So I would like to suggest all my friends to not use semicolon, which is a _JS Standard Code Style_.
[JS Standard Coding Style](https://standardjs.com/rules.html#semicolons)

> No semicolons.

lijiarui:
@Gcaufy @dcsan @zixia 
At last, I choose `mentioned()` and change the name in this commit.

src/message.ts
xinbenlv: I see. Sounds good~

JasLin:
@lijiarui @zixia  hi guys , so long a conversation...
it seems we make things complicated,  wechat identify @ with following a special char which unicode code is ``` 8197``` .  you can use ```content.charAt(n)``` to check this, ```n is  the  char position following @username ``` . 

we can rebuild this char by ```const at = String.fromCharCode(8197)``` 
so we can use ```at ``` to identify the real@.

zixia:
@JasLin awesome!

I have just remembered this, maybe someone has talked this magic Unicode number before.

xinbenlv:
Wow, so then for a copy-and-paste, will the magic unicode be carried again
in the pasted new text?


Zainan Victor Zhou, 周载南
Senior Software Engineer, Tech Lead at Google
650-336-5691

JasLin:
I just test it on Mac OS. It’s carried again.

Generally speaking, OS should not data on Simple copy & paste, unless we using formatting copy & paste. 

> 在 2017年4月14日，上午1:33，Zainan Victor Zhou <notifications@github.com <mailto:notifications@github.com>> 写道：
> 
> Wow, so then for a copy-and-paste, will the magic unicode be carried again
> in the pasted new text?
> 
> 
> Zainan Victor Zhou, 周载南
> Senior Software Engineer, Tech Lead at Google
> 650-336-5691
>

JasLin:
@xinbenlv @Gcaufy  guys，could you help to review the band feature for me pls，it's simple ，
won't spent much time：）

https://github.com/Chatie/wechaty/pull/404

src/message.ts
```diff
@@ -414,6 +414,60 @@ export class Message implements Sayable {
...
+   * ```
+   */
+  public mentioned(): Contact[] {
+    let contactList: Contact[] = []
+    const room = this.room()
+    if (this.type() !== MsgType.TEXT || !room ) {
+      return contactList
+    }
+
+    const atStringList: string[]|null = this.content().match(/@\S+ ?/g)
```
lijiarui: cool! Thanks! let me try

lijiarui:
@JasLin @xinbenlv 
> so then for a copy-and-paste, will the magic unicode be carried again
in the pasted new text

I just tested, android **won't** carried again in the pasted next text.

Android OS; MIUI 7 6.6.3 | Android 4.4.2 KOT49H

lijiarui:
@JasLin 
btw, I'm trying to use `room.say('@李佳芮' + String.fromCharCode(8197) + 'hello ')` to mock mention, but phone identify this.

lijiarui:
@JasLin @xinbenlv 
this commit has handled contact name with `@`, like `lijiarui@beijing@wechaty`

besides this, @xinbenlv and I tested and get the following conclusion: 

* **chrome web** wechat **doesn't have** the function about  red tips  `[You were mentioned]`
* **client-side** wechat **have** the function about red tips `[You were mentioned]`, but it **cannot identify** the following cases:

  * two contacts with the same name(or roomAlias)
  * copy paste mention message

*  **ios&android** wechat **have** the function about red tips `[You were mentioned]`, and **can identify** the following cases:

   * two contacts with the same name(or roomAlias)
   * copy paste mention message

zixia:
@lijiarui Could you write a summary of all the details about the `mentioned` feature to the issue or a new wiki page?

I'd like to see:

1. how to identify the magic number
1. what's the different behavior between the platforms(android/ios/web)

You could also consider writing a blog post about this story too.

src/message.ts
```diff
@@ -414,6 +414,67 @@ export class Message implements Sayable {
...
+   * ```
+   */
+  public mentioned(): Contact[] {
+    let contactList: Contact[] = []
+    const room = this.room()
+    if (this.type() !== MsgType.TEXT || !room ) {
+      return contactList
+    }
+
+    const atStrings = this.content().split(String.fromCharCode(8197))
```
lijiarui: trying to split array with mafic code `8197` and get strings atStrings.

src/message.ts
```diff
@@ -414,6 +414,67 @@ export class Message implements Sayable {
...
+        for (let i = 1; i < list.length; i++) {
+          let nick = ''
+          for (let j = 1; j <= i; j++) {
+            nick = '@' + list[list.length - j] + nick
+          }
+          nickList.push(nick.slice(1))
+        }
+        return nickList
+      }
+    })
```
lijiarui: 5. here to deal with multiple `@` string, the for cyclic to get the continuous  `@` string end with the `list` last element. for example, `e = 'Let me introduce @lijiarui@wechaty@beijing'`, then `list = ['Let me introduce ','@lijiarui','@wechaty','@beijing']`, after for cyclic, we get `nickList = ['beijing','wechaty@beijing','lijiarui@wechaty@beijing']`

lijiarui:
about message mention event:

|  | Web  |  Mac PC Client | iOS Mobile |  android Mobile |
| -------- |--------|--------|--------|--------|
| [You were mentioned] tip ([有人@我]提示)   | ✘  |    √ | √ | √ |
| Identify magic code (8197) by copy & paste in mobile  | ✘  |  √  | √   | ✘ |
| Identify magic code (8197) by programming    | ✘  |  ✘  |  ✘  | ✘ |
| Identify two contacts with the same roomAlias by [You were  mentioned] tip   | ✘   | ✘   | √ | √ |

src/message.ts
JasLin: I will approve it, but this part is a little hard to read...:)

zixia:
Thanks all for approving this PR.

@lijiarui Could you please add comments to document the function? I want you to at least include the following part:

1. your `message mention event table` and
1. your explaination(reply/review) of your PR

Because your code is still hard to read/understand, it will be better with the comments.

lijiarui:
@zixia Could you merge this time or give some suggestion?
thanks

src/message.ts
```diff
@@ -414,6 +414,93 @@ export class Message implements Sayable {
...
+    .filter(e => !!e) // filter blank string
+    const mentionList = strings.reduce((acc, val) => {
+      return acc.concat(val)
+    }, [])
+
+    log.verbose('Message', 'mentioned(%s),get mentionList: %s', this.content(), JSON.stringify(mentionList))
+    mentionList.forEach(name => {
```
zixia: I think we should use this line to replace the 15 lines forEach function:
`contactList = mentionList.map(memberAll.bind(room)).filter(c => !!c)`

src/message.ts
```diff
@@ -414,6 +414,93 @@ export class Message implements Sayable {
...
+    let contactList: Contact[] = []
+    const room = this.room()
+    if (this.type() !== MsgType.TEXT || !room ) {
+      return contactList
+    }
+
+    // Example: `Let me introduce @lijiarui@wechaty@beijing , she is a wechaty contributor`
+    // Trying to split array with magic code `8197` and get atStrings
+    // Example result: atStrings = ['Let me introduce @lijiarui@wechaty@beijing', ', she is a wechaty contributor']
+    const atStrings = this.content().split(String.fromCharCode(8197))
```
zixia: 1. Define the magic char, instead of direct use it.
1. Keep your naming style consist. use `StrList` instead of `Strings`

```ts
  const AT_SEPRATOR = String.fromCharCode(8197)
  const atStrList = this.content().split(AT_SEPRATOR)
```

src/message.ts
```diff
@@ -414,6 +414,93 @@ export class Message implements Sayable {
...
+
+    if (atStrings.length === 0) return contactList
+
+    // Using `filter(e => e.indexOf('@') > -1)` to filter the string without `@`
+    // Example result: atStrings.filter(e => e.indexOf('@') > -1) = ['Let me introduce @lijiarui@wechaty@beijing']
+    const strings = atStrings.filter(e => e.indexOf('@') > -1).map(e => {
```
zixia: 1. Do not use var name `strings` because it will be very confusing with `string`
1. Do not use a map inline function with so many lines.
1. Do not use `e` as the var name. Because `e` stands for `event`, `error`, etc. use `s`/`str` for the string instead.

I can not understand the code inside this map function. 

However, I would like to rewrite this huge map function like this:

```ts
const mentionedStringList = atStrList
  .filter(str => str.include('@))
  .map(str => str.replace(/^.*?@/, '@'))
```

src/message.ts
```diff
@@ -414,6 +414,93 @@ export class Message implements Sayable {
...
+          }
+          nickList.push(nick.slice(1))
+        }
+
+        // Example result: nickList = ['beijing','wechaty@beijing','lijiarui@wechaty@beijing']
+        return nickList
+      }
+    })
+    .filter(e => !!e) // filter blank string
+    const mentionList = strings.reduce((acc, val) => {
```
zixia: What's this `reduce` function meaning for?

I feel the mentionList will be equal to your `strings` array.

src/message.ts
```diff
@@ -414,6 +414,93 @@ export class Message implements Sayable {
...
+        return [e.slice(e.lastIndexOf('@') + 1)]
+      } else {
+
+        // Deal with contact with multiple `@`signs
+        // Select element from `list` array to join strings with the following rules:
+        // 1. end with the last element from `list` array.
+        // 2. element are continuous, eg: [3],[2,3],[1,2,3]
+
+        // Here: list = ['Let me introduce ','@lijiarui', '@wechaty', '@beijing']
+        let nickList: string[] = []
```
zixia:
```js
const str = '@a@b@c'

let tryName = ''
const tryNameList = []

str.split('@')
  .filter(c => !!c)
  .forEach(c => {
    tryName += '@' + c
    tryNameList.push(tryName.slice(1)) // get rid of the `@` at beginning
  })

console.log(tryNameList)
```

will output: `[ 'a', 'a@b', 'a@b@c' ]`

Please use the above code to replace your for loop.

lijiarui:
I'll pr `room.say()` by magic code, but as mention table showed, it won't get red [You were mentioned] tip on mobile wechat. While anyway, it should change.

zixia:
Awesome, thanks!

PR #364, opened 2017-03-27
room.memberAll() & change room.member() query to 3 types

lijiarui:
#334 

- add function `room.memberAll():Contact[]` correspond to `room.member(): Contact`
- change room.member() query to 3 types:
  - name
  - roomAlias
  - contactAlias

PR #370, opened 2017-03-28
add room-leave event

### LINKED ISSUE #250: add `room-bot-leave` event
We have a `room-event` when bot removed others from the room, but when others removed the  bot from the room, we don't have any event.

So I suggest adding an event when the bot is removed from the owner of one room.

Then I have a question: 

##### 2. `room-leave`: use `leaverList` to identify the differences.

I prefer the `room-bot-leave`,  because I think `bot remove others from the room` and `others remove bot from the room` used in different scene.

#### Sys Message as follows:

> You were removed from the group chat by "桔小秘"
> 你被"李佳芮"移出群聊

lijiarui:
fix #250


zixia:
Sorry, i cannot merge this because your PR didn't pass any of CI test.

lijiarui:
Yes, it cannot be merge now, because I should fix #364 first.

And I have fix room.member(), please help to merge it first, thanks

zixia:
Sorry, I have no time to help you about this recently. You need to do it by yourself.

lijiarui:
@zixia maybe it has done...

src/puppet-web/firer.ts
```diff
@@ -315,22 +324,35 @@ async function checkRoomLeave(m: Message): Promise<void> {
...
    */
-  let leaverContact = room.member(leaver)
-
-  if (!leaverContact) {
-    log.error('PuppetWebFirer', 'fireRoomLeave() leaver %s not found, event `room-leave` & `leave` will not be fired', leaver)
-    return
+  let leaverContact: Contact | null, removerContact: Contact | null
+  if (leaver === this.userId) {
+    leaverContact = Contact.load(this.userId)
+    removerContact = room.member({alias: remover}) || room.member({name: remover})
```
zixia: Please use the formal query filter name, instead of using an alias.

```ts
removerContact = room.member({roomAlias: remover}) || room.member({name: remover})
```

I will more prefer to write this piece of code in this way:
```ts
removerContact = room.member(remover)
```

Because `Room.member()` should do the job for translate the string name to a Contact inside a room.

lijiarui: I think using `room.member()` is not strict here， because `roomAlias`(alias) here is useless, bot just recognize `contactAlias` and `name`, if we use `roomAlias`, then when someone's roomAlias is the same as other's name in the room. it will get the contact we don't mean go get.

So, maybe the new change I commit is the better way. we should may the code easier for developer, but more strict here.

src/puppet-web/firer.ts
```diff
@@ -315,22 +324,35 @@ async function checkRoomLeave(m: Message): Promise<void> {
...
+  if (leaver === this.userId) {
+    leaverContact = Contact.load(this.userId)
+    removerContact = room.member({contactAlias: remover}) || room.member({name: remover})
+    if (!removerContact) {
+      log.error('PuppetWebFirer', 'fireRoomLeave() bot is removed from the room, but remover %s not found, event `room-leave` & `leave` will not be fired', remover)
+      return
+    }
+  } else {
+    removerContact = Contact.load(this.userId)
+    leaverContact = room.member({contactAlias: remover}) || room.member({name: leaver})
```
zixia: I think it is better to use `member(name)`  here for better encapsulation.

zixia:
Please get 2 more approving from other contributors as well, then I could be able to merge this. 

Thanks!

zixia:
@lijiarui This is the way of my favorite for using `room.member()`, which have already used many times in the code base:

https://github.com/Chatie/wechaty/blob/master/src/puppet-web/firer.ts#L319

zixia:
The shorter, the better.

zixia:
Thanks all! 

I can merge this PR now, Cheers!

PR #376, opened 2017-03-30
Limit the size of the sending file

mukaiu:
Wechat web limits the file size to 20M

### DIFF
--- a/src/puppet-web/puppet-web.ts
+++ b/src/puppet-web/puppet-web.ts
@@ -353,9 +353,9 @@ export class PuppetWeb extends Puppet {
       }))
     })
 
-    let fileMaxSize = 20 * 1024 * 1024
-    if (buffer.length > fileMaxSize)
-      throw new Error('Maximum file size 20M')
+    const videoMaxSize = 20 * 1024 * 1024
+    if (mediatype == 'video' && buffer.length > videoMaxSize)
+      throw new Error('Sending video files is not allowed to exceed 20MB')
 
     let md5 = UtilLib.md5(buffer)
 
--- a/src/puppet-web/puppet-web.ts
+++ b/src/puppet-web/puppet-web.ts
@@ -357,7 +357,7 @@ export class PuppetWeb extends Puppet {
     // https://github.com/Chatie/webwx-app-tracker/blob/7c59d35c6ea0cff38426a4c5c912a086c4c512b2/formatted/webwxApp.js#L1115
     const videoMaxSize = 20 * 1024 * 1024
     if (mediatype === 'video' && buffer.length > videoMaxSize)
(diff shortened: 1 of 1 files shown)


src/puppet-web/puppet-web.ts
```diff
@@ -353,6 +353,10 @@ export class PuppetWeb extends Puppet {
       }))
     })
 
+    let fileMaxSize = 20 * 1024 * 1024
```
zixia: 1. Should be const instead of let.
2. Could you link to the size limit code line in wxwebapp(we have a track tool named webwxapp-tracker
)? So we could track it in the future.

src/puppet-web/puppet-web.ts
```diff
@@ -353,6 +353,10 @@ export class PuppetWeb extends Puppet {
...
+    if (buffer.length > fileMaxSize)
+      throw new Error('Maximum file size 20M')
```
zixia: 20M should be calculated here instead of hard coded.

src/puppet-web/puppet-web.ts
```diff
@@ -353,6 +353,12 @@ export class PuppetWeb extends Puppet {
       }))
     })
 
+    // Sending video files is not allowed to exceed 20MB
+    // https://github.com/Chatie/webwx-app-tracker/blob/master/formatted/webwxApp.js#L1115
```
zixia: Please use the link with a specific commit hash number instead of the latest, because what we want to track is a specific version.

like this: <https://github.com/Chatie/webwx-app-tracker/blob/7c59d35c6ea0cff38426a4c5c912a086c4c512b2/formatted/webwxApp.js#L1115>

src/puppet-web/puppet-web.ts
```diff
@@ -353,6 +353,12 @@ export class PuppetWeb extends Puppet {
...
+    const videoMaxSize = 20 * 1024 * 1024
+    if (mediatype === 'video' && buffer.length > videoMaxSize)
+      throw new Error('Sending video files is not allowed to exceed 20MB')
```
zixia: `     throw new Error('Sending video files is not allowed to exceed ' + videoMaxSize/1024/1024 + 'MB')`

mukaiu: OK

mukaiu:
My code details need to be optimized😄😄😄

src/puppet-web/puppet-web.ts
mukaiu: using 
```
`Sending video files is not allowed to exceed ${videoMaxSize/1024/1024}MB`
```

zixia:
Great, thanks!

Issue #378, opened 2017-03-30
jsdoc2md may flush some pieces of the embedded doc


ax4:
**In brief, this issue is caused by our auto-doc working flow, at the 1st step `npm run dist` (compile TS into JS), it flushes some pieces of embedded doc** May Impact #252 #73 

Our auto-doc working flow (with **jsdoc / jsdoc2md**):
```md
1. dev in TS
2. embedded doc insert in TS
3. compile TS into JS (#warning# this step cause flush)
4. run jsdoc / jsdoc2md
5. get the final doc
```

**See more in details:** 
1. A piece of embedded doc in TS 
https://github.com/Chatie/wechaty/blob/b77c241d7d716635c5481ed734b32e303fce8e4c/src/contact.ts#L430-L505
2. The compiled JS missing "that piece"
![flush_cause](https://cloud.githubusercontent.com/assets/8946060/24517748/0082426c-15b2-11e7-9637-23908881229f.png)
3. Cause the missing in the final doc
![flush_result](https://cloud.githubusercontent.com/assets/8946060/24517754/057927ea-15b2-11e7-84f9-5b95605ab9f1.png)

ax4:
I will look deeper into this problem, and **trying to solve it, within 24 hours!**  

**Possible solution: (I haven't tried, I will do it)** 
1. Give up the ['one-to-one edit strategy' like this](https://github.com/Chatie/wechaty/blob/b77c241d7d716635c5481ed734b32e303fce8e4c/src/contact.ts#L430-L505), try to put only one piece for the same name function, and see if this can avoid flush?
2. Use TypeDoc, it generate good & informative doc, directly from TS files
3. Manually apply the missing piece onto JS files (need human / bot to do the job)

zixia:
Glad to see you start working on this! :+1:

zixia:
Merge to #252

PR #380, opened 2017-03-31
fix jsdoc flush issue #378 and minor fix on the doc examples

ax4:
#378 I have fixed this flush issue! 

And also, I made 2 minor fix: 
* Fix 'set' => 'Set' of upper case.  
* Fix examples on contact.alias() function.  User '.alias()' instead of '.remark // should be deprecated '

### DIFF
--- a/src/contact.ts
+++ b/src/contact.ts
@@ -428,7 +428,7 @@ export class Contact implements Sayable {
   }
 
   /**
-   * Get the alias for contact
+   * GET the alias for contact
    *
    * @returns {(string | null)}
    *
@@ -440,7 +440,7 @@ export class Contact implements Sayable {
   public alias(): string | null
 
   /**
-   * Set the alias for contact
+   * SET the alias for contact
    *
    * tests show it will failed if set alias too frequently(60 times in one minute).
    *
@@ -460,7 +460,7 @@ export class Contact implements Sayable {
   public alias(newAlias: string): Promise<boolean>
 
   /**
-   * Delete the alias for a contact
+   * DELETE the alias for a contact
    *
    * @param {null} empty
    * @returns {Promise<boolean>}
--- a/src/contact.ts
+++ b/src/contact.ts
@@ -485,7 +485,12 @@ export class Contact implements Sayable {
    *
    * @example GET the alias for a contact
    * ```ts
-   * const alias = contact.alias()  //@returns {(string | null)}
+   * const alias = contact.alias()
+   * if (alias === null) {
+   *   console.log('You have not yet set any alias for contact ' + contact.name())
+   * } else {
+   *   console.log('You have already set an alias for contact ' + contact.name() + ':' + alias)
+   * }
    * ```
    *
    * @example SET the alias for a contact
--- a/src/contact.ts
+++ b/src/contact.ts
@@ -440,7 +440,7 @@ export class Contact implements Sayable {
   public alias(): string | null
 
   /**
-   * set the alias for contact
+   * Set the alias for contact
    *
    * tests show it will failed if set alias too frequently(60 times in one minute).
    *
--- a/src/contact.ts
+++ b/src/contact.ts
@@ -469,9 +469,9 @@ export class Contact implements Sayable {
    * ```ts
    * const ret = await contact.alias(null)
    * if (ret) {
-   *   console.log('ok!')
+   *   console.log(`delete ${contact.name()}'s alias successfully!`)
    * } else {
-   *   console.error('fail!')
+   *   console.log(`failed to delete ${contact.name()}'s alias!`)
    * }
    * ```
    */
@@ -507,9 +507,9 @@ export class Contact implements Sayable {
    * ```ts
    * const ret = await contact.alias(null)
    * if (ret) {
-   *   console.log('ok!')
+   *   console.log(`delete ${contact.name()}'s alias successfully!`)
    * } else {
-   *   console.error('fail!')
+   *   console.log(`failed to delete ${contact.name()}'s alias!`)
    * }
    * ```
    */
--- a/src/contact.ts
+++ b/src/contact.ts
@@ -477,6 +477,37 @@ export class Contact implements Sayable {
    */
   public alias(empty: null): Promise<boolean>
 
+  /**
+   * GET / SET / DELETE the alias for a contact
+   *
+   * @param {(none | string | null)} newAlias , 
+   * @returns {(string | null | Promise<boolean>)}
+   *
+   * @example GET the alias for a contact
+   * ```ts
+   * const alias = contact.alias()  //@returns {(string | null)}
+   * ```
+   * 
+   * @example SET the alias for a contact
+   * ```ts
+   * const ret = await contact.alias('lijiarui')  //@returns {Promise<boolean>}
+   * if (ret) {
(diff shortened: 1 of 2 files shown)


src/contact.ts
```diff
@@ -477,6 +477,37 @@ export class Contact implements Sayable {
...
 
+  /**
+   * GET / SET / DELETE the alias for a contact
+   *
+   * @param {(none | string | null)} newAlias , 
+   * @returns {(string | null | Promise<boolean>)}
+   *
+   * @example GET the alias for a contact
+   * ```ts
+   * const alias = contact.alias()  //@returns {(string | null)}
```
zixia: Please use code to talk instead of comments, especially in examples section.

This will be better:

```ts
const alias = contact.alias()
if (alias === null) {
  console.log('You did not set any alias for contact ' + contact.name())
} else {
  console.log('You had set alias for contact ' + contact.name() + ':' + alias)
}
```

src/contact.ts
```diff
@@ -477,6 +477,37 @@ export class Contact implements Sayable {
...
+   * ```
+   * 
+   * @example SET the alias for a contact
+   * ```ts
+   * const ret = await contact.alias('lijiarui')  //@returns {Promise<boolean>}
```
zixia: This one is splendid because of your example show code instead of comments.

You can get rid of comments because the code can tell everything.

src/contact.ts
```diff
@@ -477,6 +477,37 @@ export class Contact implements Sayable {
...
+   * if (ret) {
+   *   console.log(`change ${contact.name()}'s alias successfully!`)
+   * } else {
+   *   console.error('failed to change ${contact.name()}'s alias!')
+   * }
+   * ```
+   * 
+   * @example DELETE the alias for a contact
+   * ```ts
+   * const ret = await contact.alias(null)  //@returns {Promise<boolean>}
```
zixia: Get rid of the comments in example code too.

Example code should be very easy to understand without any comments. Or there's must be something wrong.

src/contact.ts
```diff
@@ -440,7 +440,7 @@ export class Contact implements Sayable {
   public alias(): string | null
 
   /**
-   * set the alias for contact
+   * Set the alias for contact
```
lijiarui: If you use `GET / SET / DELETE the alias for a contact` on line 481, then I suggest here should be `SET`  for unification.
Or, change line 481 to `Get / Set / Delete the alias for a contact`.

It should change as well as `Set` and `Delete`

You decide it.

src/contact.ts
ax4: I got it. Thanks! Modifying

src/contact.ts
```diff
@@ -477,6 +477,37 @@ export class Contact implements Sayable {
...
+   * } else {
+   *   console.error('failed to change ${contact.name()}'s alias!')
+   * }
+   * ```
+   *
+   * @example DELETE the alias for a contact
+   * ```ts
+   * const ret = await contact.alias(null)  //@returns {Promise<boolean>}
+   * if (ret) {
+   *   console.log('ok!')
```
lijiarui: Maybe here should changed as follows, and it would be better than I wrote before.
```ts
const ret = await contact.alias(null)
if (ret) {
  console.log(`delete ${contact.name()}'s alias successfully!`)
} else {
  console.log(`failed to delete ${contact.name()}'s alias!`)
}
```

src/contact.ts
```diff
@@ -477,6 +477,37 @@ export class Contact implements Sayable {
...
+   *
+   * @example DELETE the alias for a contact
+   * ```ts
+   * const ret = await contact.alias(null)
+   * if (ret) {
+   *   console.log('ok!')
+   * } else {
+   *   console.error('fail!')
+   * }
+   * ```
```
lijiarui: Maybe here should be changed as follows, it would better than I wrote before:
```ts
const ret = await contact.alias(null)
if (ret) {
  console.log(`delete ${contact.name()}'s alias successfully!`)
} else {
  console.log(`failed to delete ${contact.name()}'s alias!`)
}
```

src/contact.ts
```diff
@@ -477,6 +477,37 @@ export class Contact implements Sayable {
...
+   * const alias = contact.alias()  //@returns {(string | null)}
+   * ```
+   *
+   * @example SET the alias for a contact
+   * ```ts
+   * const ret = await contact.alias('lijiarui')
+   * if (ret) {
+   *   console.log(`change ${contact.name()}'s alias successfully!`)
+   * } else {
+   *   console.error('failed to change ${contact.name()}'s alias!')
```
lijiarui: You should change **'** to **`** here, or it may cannot be parsed successfully.
```ts
console.error(`failed to change ${contact.name()}'s alias!`)
```

src/contact.ts
```diff
@@ -449,7 +449,7 @@ export class Contact implements Sayable {
    *
    * @example
    * ```ts
-   * const ret = await contact.remark('lijiarui')
+   * const ret = await contact.alias('lijiarui')
    * if (ret) {
    *   console.log(`change ${contact.name()}'s alias successfully!`)
    * } else {
```
lijiarui: I was wrong here too... Please help me to fix
```ts
console.error(`failed to change ${contact.name()}'s alias!`
```

src/contact.ts
ax4: Maybe we should unify the ' or ` issue. 

I see a lot of ' , while having a lot of ` too. 

On my laptop, they could both parsed successfully. with `npm run doc` and 'npm run lint' is okay.

src/contact.ts
ax4: Maybe we should unify the ' or ` issue.

I see a lot of ' , while having a lot of ` too.

On my laptop, they could both parsed successfully. with npm run doc and 'npm run lint' is okay.

src/contact.ts
ax4: Ok I got it! I would unify them into GET / SET / DELETE

src/contact.ts
ax4: fixed

src/contact.ts
ax4: fixed

src/contact.ts
ax4: fixed.

src/contact.ts
ax4: Ok, Modifying .

src/contact.ts
lijiarui: maybe ` is better, because sometimes there is `'` in the log, it will cut off the log.

src/contact.ts
ax4: fixed.

src/contact.ts
ax4: fixed.

src/contact.ts
ax4: Ok, I've setup issue #381 to fix the entire thing, tomorrow!

lijiarui:
@ax4 
One last thing, `src/puppet-web/puppet-web.ts` is this change by you or others?
Maybe this file shouldn't be committed here.

src/contact.ts
ax4: resolved! Will fix the entire ' or ` issue at #381

zixia:
@ax4 @lijiarui Thanks for the PR and the approving!

PR #404, opened 2017-04-12
support brand checking of contact 

JasLin:
```
@see https://github.com/Urinx/WeixinBot/blob/master/README.md

公众号/服务号	以@开头，但其VerifyFlag & 8 != 0 

VerifyFlag: 
         一般公众号/服务号：8 
         微信自家的服务号：24 
         微信官方账号微信团队：56
```


tslint.json
```diff
@@ -36,7 +36,7 @@
     "member-ordering": [false],
     "no-any": false,
     "no-arg": true,
-    "no-bitwise": true,
```
zixia: I'd like to keep this because the bitwise operation is very rare in our code.

src/contact.ts
```diff
@@ -110,6 +112,11 @@ export class Contact implements Sayable {
       star:       !!rawObj.StarFriend,
       stranger:   !!rawObj.stranger, // assign by injectio.js
       avatar:     rawObj.HeadImgUrl,
+      /**
+       * @see 1. https://github.com/nodeWechat/wechat4u/blob/master/src/interface/contact.js#L65
+       * @see 2. https://github.com/Urinx/WeixinBot/blob/master/README.md
+       */
+      brand:      !!rawObj.UserName && !rawObj.UserName.startsWith('@@') && (rawObj.VerifyFlag & 8) !== 0,
```
zixia: I'd like to use this line to disdable tslint temporary
```ts
// tslint:disable-next-line
```

src/contact.ts
```diff
@@ -153,6 +160,21 @@ export class Contact implements Sayable {
...
+   *
+   * @returns {boolean|null} True for brand, Flase for contact is not a brand, null for can not get the info
+   *
+   * @example
+   * ```ts
+   * const isBrand = contact.brand()
+   * ```
+   */
+  public brand(): boolean|null {
+    if (!this.obj) return null
```
zixia: Our framework should always to make sure a `contact` is `ready`, which means `this.obj` should not be empty.

If so, it's more like a bug and should throw an exception.

So my suggestion is to change this function to:

```ts
public brand(): boolean {
  return this.obj && this.obj.brand
}
```

example/contact-bot.ts
```diff
@@ -78,6 +78,20 @@ async function main() {
   log.info('Bot', '#######################')
   log.info('Bot', 'Contact number: %d\n', contactList.length)
 
+  /**
+   * List Brand Contact
+   */
+  for (let i = 0; i < contactList.length; i++) {
+    const contact = contactList[i]
+    if (contact.brand() === null) {
```
zixia: The `contact` here should already `ready()`-ed, because Wechaty should make sure it is.

If it is not, then Wechaty must have a bug in somewhere...

src/contact.ts
JasLin:
```
// tslint:disable-next-line
```

will disables all rules for the following lines, it seems it's not a good choose for this case.

how about this?

```ts
!!rawObj.UserName && !rawObj.UserName.startsWith('@@') && [8, 24, 56].indexOf(rawObj.VerifyFlag) > -1,
```

src/contact.ts
JasLin:
```ts
public brand(): boolean {
  return !!this.obj && this.obj.brand
}
```

JasLin:
@lijiarui @Gcaufy @xinbenlv @dcsan @imWildCat @mukaiu 
hi guys , anybody have time to review this feature?

tslint.json
JasLin: yes, u r right:)

src/contact.ts
zixia: I thought that will only disable the following ONE line?

Could you please have a check about it, because the new version of your code is difficult to read...

src/contact.ts
zixia: Yes your code is correct. :)

src/contact.ts
JasLin: yes, ...next-line... u r right:)

JasLin:
@lijiarui thanks , i am trying to review them

lijiarui:
@dcsan I guess maybe you need this feature most, could you help to review this please? Thanks

dcsan:
@lijiarui I'm not sure what this feature does?

> +   * Check if it's a brand contact

what is a `brand` in this context? It usually means a product brand. Maybe you mean `brand new` ?
Sorry but I'm unclear.

lijiarui:
@dcsan `brand` here means `wechat oa account`

Generally ,wechaty cannot identify the contact is a real contact or a wechat oa account, but this pr can fix this problem.

dcsan:
ok mingbai! so this is in the case where chatie bot is following an official account. 
I'm not clear exactly what the use case would be for that - maybe testing other OAs?

so we have:
- `msg.room() ` - to see if it is a chat GROUP or a 1:1 (DM) chat
- `msg.brand()` - to see if it's an OA or ( a normal user OR group )

then service accounts, subscription accounts, personal OAs - not sure if the wechaty client gets any information about these.

`msg.contact()` - isn't really relevant for groups. maybe `msg.sender()`

then we could have a `msg.sender().type()` that returns
- `brand | user | group`

as a string?

If we use enums btw I think we need to make these more easily exportable/importable to work with JS. Now we have `async/await` in node7 I've switched back to JS myself for most new projects.

zixia:
@dcsan IMO stay with TypeScript is the bright decision, I love it to death now. :P

And for this `VerifyFlag` I think the most useful part is to identify if a `contact` is a personal account.

Now we can get all contact after login by `Contact.findAll()`, but the contactList is a mix of _personal users_ and _official accounts_.

Add a `contact.official()` could sperate them easily by:

```ts
if (contact.official()) {
  // is Official Account
} else {
  // is Personal Account
}
```

dcsan:
I think it's also useful to know if its a group(room) or an individual user, as some of the behavior is different, eg I don't think there is a `contact` for a room? (and not sure if there is a `msg.room()` or a `topic()` for an individual user. So you have to be careful with the logic or code can break, the types are not polymorphic.

I think i have stuff like:
```
topic = msg.room() ? msg.room().topic() : 'dm'
```

So when you have three major possible types, a boolean check method isn't as useful?

----
Personally I like Typescript a lot, but it is hard just getting new developers up and running, especially we have some team in Russia and other places. typescript, tslint, wechat api, docker... hours of setup to even get to helloworld.

zixia:
Yes, I agree with you that TypeScript has a huge study curve. It seems things getting better after TypeScript v2.3 because they became treating unknown things to be `any` by default instead of throw errors, which will make JS developer easier to getting start.

And I do not fully understand you, as in the reply you said: 
>  to know if its a group(room) or an individual user

To clarify, In Wechaty:
1. room(group) will be an instance of `Room` and,
1. contact(individual user) will be an instance of `Contact`. Room and Contact are totally two different Class(abstract);
1. There is no contact for a room. Contact is for one, Room is for many;
1. There is a `message.room(): Room` method to identify which room this message comes from;
1. There is no `contact.topic()` because that's meaningless;

For your code:
> topic = msg.room() ? msg.room().topic() : 'dm'

I'd like to rewrite it to:

```ts
const room = msg.room()
let topic: string
if (room) { 
  // this message is received in a room
  assert(room instanceof Room)
  topic = room.topic()
} else {
  // this message is received from a indivial user

  // WHY do you set room topic 
  // when you are receiving a direct message 
  // from an individual user?
  topic = 'dm' // ???
}

```

dcsan:
hi - thanks for the above. 

```
    topic = msg.room() ? msg.room().topic() : 'dm'
    // WHY do you set room topic 
    // when you are receiving a direct message 
    // from an individual user?
```

yeah, it's a bit hacky agreed. it is so we can render all messages and handle group|dm display based on a single (cached) property rather than another chain of method calls.
But they do still have a different class hierarchy so it's not that correct. 
Now we have to check two methods:

`if ( msg.room() === null && msg.brand() === null ) ` then it's a `DM`

If we had `msg.type()` we could do:

```js
switch (msg.type() ):
   case 'room':
      // responding in a group
   case 'dm':
     //..
   case 'oa':
    // brand / official account
```

Just a thought. This maybe still a hacky way to do this, as they are still different types of objects. Perhaps we'd be better with something like the `Message < MediaMessage` hierarchy.

----
re TypeScript, I think the language is easy.
it's all the setup of the tooling, linter, compiler (global tsc? compile output to where? .tsconfig settings?) especially for new apps. We are trying to break out more into microservices now, so we have a few smaller apps. this creates more setup overhead with TS.

zixia:
@JasLin Hi bro, 

I suggest:

1. we rename `brand()` to `official()` and
1. include all the VerifyFlag(8/24/56) and
1. launch this PR for now. 

Because we could use this method to identify whether the contact is an individual personal user.

How do you think?

JasLin:
@zixia sorry check this so later，```brand()``` is come from wxwebapp source code ，but i love ```official()``` too.

I will try to finish this tomorrow. And, happy birthday bro!:)

JasLin:
@zixia Hi bro,  I just check the wxwebapp's source and the rawObj again . the following info is confirmed

* we don't need to include all verifyFlag , use ```!!(VerifyFlag & 8)``` can identify all brand/offical account , check [isBrandContact](https://github.com/Chatie/wechaty/blob/master/doc/webwxapp.js#L3368)  method from source 
* VerifyFlag should be define like this exactly: 


> VerifyFlag: 
        ~~一般公众号/服务号：8~~  个人公众号: 8
         ~~微信自家的服务号：24~~  企业公众号： 24
         微信官方账号微信团队：56

and two more thing : 

* should we add a method ```personal() {return !official()}```to identify personal() account for convenient?
* should we add ```special()``` to identify special account ? check the source [isSpContact](https://github.com/Chatie/wechaty/blob/master/doc/webwxapp.js#L4187)

zixia:
```
VerifyFlag:
个人公众号: 8
企业公众号： 24
微信官方账号微信团队：56
```

I think we could translate them into bit position:

| Type | 10 | 2 | Bit |
|--- | ----|  ----| ---|
| Official Account | 8 | 0000 1000 | 1 << 4 |
| Business Verified | 24 | 0001 1000 | 1 << 5 |
| Tencent Verified | 56 | 0011 1000 | 1 << 6 |

How about let's redesign `official()` like this:

```ts
enum OfficialType {
  OFFICIAL,
  BUSINESS,
  TENCENT,
}

official(type = OfficialType.OFFICIAL): boolean
```

By default, `official()` will return whether the contact is an official account(1<<4 bit set). If we want to know whether it's a business verified, we could call `official(OfficialType.BUSINESS)` or `official(OfficialType.TENCENT)`, and we could extend to more types for the official account if needed in the future.

* Abut the `personal()` method, I have no feeling about it. :P
* About the `isSpContact()`, I have no idea about what it is. I also see other names `isShieldContact()` etc, I think we could talk about them in another issue?

src/contact.ts
```diff
@@ -61,6 +64,16 @@ export type ContactQueryFilter = {
 }
 
 /**
+ * @see https://github.com/Chatie/wechaty/blob/master/doc/webwxapp.js#L4057
```
zixia: Please do not use `wechaty/doc`, because this directory is deprecated.

Should change refer to <https://github.com/Chatie/webwx-app-tracker> with the exactly commit hash.

src/contact.ts
```diff
@@ -110,6 +123,16 @@ export class Contact implements Sayable {
       star:       !!rawObj.StarFriend,
       stranger:   !!rawObj.stranger, // assign by injectio.js
       avatar:     rawObj.HeadImgUrl,
+      /**
+       * @see 1. https://github.com/Chatie/wechaty/blob/master/doc/webwxapp.js#L3368
```
zixia: Please change refer to https://github.com/Chatie/webwx-app-tracker with the exactly commit hash.

src/contact.ts
```diff
@@ -110,6 +123,16 @@ export class Contact implements Sayable {
...
+       * @see 2. https://github.com/Urinx/WeixinBot/blob/master/README.md
+       */
+      // tslint:disable-next-line
+      official:      !!rawObj.UserName && !rawObj.UserName.startsWith('@@') && !!(rawObj.VerifyFlag & 8),
+      /**
+       * @see 1. https://github.com/Chatie/wechaty/blob/master/doc/webwxapp.js#L4187
```
zixia: Please change refer to https://github.com/Chatie/webwx-app-tracker with the exactly commit hash.

src/contact.ts
```diff
@@ -153,6 +176,58 @@ export class Contact implements Sayable {
...
+   *
+   * the contact who's id in following list will be identify as a special contact
+   *
+   * ```ts
+   * 'weibo', 'qqmail', 'fmessage', 'tmessage', 'qmessage', 'qqsync', 'floatbottle',
+   * 'lbsapp', 'shakeapp', 'medianote', 'qqfriend', 'readerapp', 'blogapp', 'facebookapp',
+   * 'masssendapp', 'meishiapp', 'feedsapp', 'voip', 'blogappweixin', 'weixin', 'brandsessionholder',
+   * 'weixinreminder', 'wxid_novlwrv3lqwv11', 'gh_22b87fa7cb3c', 'officialaccounts', 'notification_messages',
+   * ```
+   * @see https://github.com/Chatie/wechaty/blob/master/doc/webwxapp.js#L4057
```
zixia: Please change refer to https://github.com/Chatie/webwx-app-tracker with the exactly commit hash.

JasLin:
* actually , ```special``` contacts contain most of ```Tencent``` contacts , but i haven't  checked that whether all the ```special|Tencent``` contact's verifyFlag is 56, because i can not find out a way to load all those contacts to my WeChat. 
* and , check the source code of wxwebapp, there is no verifyFlag  defined as ```24``` and ,```56```
, so i guess that maybe [Urinx](https://github.com/Urinx/WeixinBot/blob/master/README.md)'s analysis about  ```24``` and ```56``` is just a ```guess``` too:) ,  as u see in the document ,he think ```微信自家的服务号``` 's flag is ```24``` , but in my bot, I found that most of  ```企业公众号```'s flag is ```24``` instead.  since that , i am not sure about this exactly.  but we can sure that using ```!!(VerifyFlag & 8)``` can identify all brand/offical contacts   

here are all the flag defined in source code: 
```ts
            CONTACTFLAG_CONTACT: 1,
            CONTACTFLAG_CHATCONTACT: 2,
            CONTACTFLAG_CHATROOMCONTACT: 4,
            CONTACTFLAG_BLACKLISTCONTACT: 8,
            CONTACTFLAG_DOMAINCONTACT: 16,
            CONTACTFLAG_HIDECONTACT: 32,
            CONTACTFLAG_FAVOURCONTACT: 64,
            CONTACTFLAG_3RDAPPCONTACT: 128,
            CONTACTFLAG_SNSBLACKLISTCONTACT: 256,
            CONTACTFLAG_NOTIFYCLOSECONTACT: 512,
            CONTACTFLAG_TOPCONTACT: 2048,
            ... ...
            MM_USERATTRVERIFYFALG_BIZ: 1
            , MM_USERATTRVERIFYFALG_FAMOUS: 2
            , MM_USERATTRVERIFYFALG_BIZ_BIG: 4
            , MM_USERATTRVERIFYFALG_BIZ_BRAND: 8
            , MM_USERATTRVERIFYFALG_BIZ_VERIFIED: 16
```
so I suggest we design ```official()``` at this time , after we find out what is ```24```, ```56``` exactly means , then we can change ```office()``` to ```official(type?): boolean```.

src/contact.ts
JasLin: had changed it with the latest commit

JasLin:
it seems we  need more contributors ...

zixia:
Yes I agree with you bro... @_@

PR #409, opened 2017-04-13
mini qrcode img

lijiarui:
qrcode img is too big, change the same size with weixinBot


zixia:
Why? 

Currently size is 430x430 which I feels good.

What's the reason to reduce it to 220x220?

PR #413, opened 2017-04-18
Fixed image upload fail

nothingisover:
If hostname == wx2.qq.com
The file host should be "file2.wx.qq.com"


zixia:
Hi @nothingisover, 

Thanks for catching the bug and send this PR!

However, it looks that @mukaiu has a better PR #415  which will be more reliable, so I'll merge #415 instead of this one.

Thank you again for your contribution, have a good day!

Issue #175, opened 2016-12-30
[ci] no new tests completed within the last 180000ms of inactivity


zixia:
PuppetWebBrowser open(https://wx.qq.com)

---

https://travis-ci.org/wechaty/wechaty/jobs/187718144#L2720

```shell
VERB PuppetWebBrowser open(https://wx.qq.com)
  ✖ Exited because no new tests completed within the last 180000ms of inactivity
  44 tests passed [15:50:57]
  1 test skipped
  1 uncaught exception
```

---

https://travis-ci.org/wechaty/wechaty/jobs/187832366#L940

```shell
VERB PuppetWebBrowser open(https://wx.qq.com)
  ✖ Exited because no new tests completed within the last 180000ms of inactivity
  43 tests passed [02:22:10]
  1 test skipped
  1 uncaught exception
```

---

https://travis-ci.org/wechaty/wechaty/jobs/187832421#L895

```shell
VERB PuppetWebBridge proxyWechaty(quit, )
  ✖ Exited because no new tests completed within the last 180000ms of inactivity
  41 tests passed [02:23:38]
  1 test skipped
  1 uncaught exception
```

---

https://travis-ci.org/wechaty/wechaty/jobs/190391290#L3696

```shell
VERB PuppetWebBrowser open(https://wx.qq.com)
SILL TestPuppetWeb waitDing() retryPromise: attampt 3/13 time for timeout 84500
VERB PuppetWebBridge ding(dingdong)
VERB PuppetWebBridge proxyWechaty(ding, dingdong)
SILL PuppetWebBridge execute()
SILL PuppetWebBrowser Browser.execute("return typeof WechatyBro === "undefined"")
SILL StateMonitor Browser:target() - open
SILL StateMonitor Browser:current() - open
  ✖ Exited because no new tests completed within the last 180000ms of inactivity
```

---

https://travis-ci.org/wechaty/wechaty/jobs/190943656#L3582

```shell
VERB PuppetWebBrowser getBrowserPidList()
SILL PuppetWebBrowser getBrowserPids() child: {"PPID":"7569","PID":"7703","STAT":"Sl+","COMMAND":"chrome"}
SILL PuppetWebBrowser getBrowserPids() child: {"PPID":"7703","PID":"7716","STAT":"S+","COMMAND":"chrome"}
SILL PuppetWebBrowser getBrowserPids() child: {"PPID":"7716","PID":"7749","STAT":"Sl+","COMMAND":"chrome"}
[...]

PR #416, opened 2017-04-18
Fix chrome driver path problem in Windows

xjchengo:
After install from souce in Windows, `npm run demo` throws:
```
21:13:29 ERR PuppetWebBrowserDriver initChromeDriver() Wechaty require `chromedriver` to be installed.(try to run: "npm
install chromedriver" to fix this issue)
21:13:29 ERR PuppetWebBrowser init() exception: The ChromeDriver could not be found on the current PATH. Please download
 the latest version of the ChromeDriver from http://chromedriver.storage.googleapis.com/index.html and ensure it can be
found on your PATH.
```

I solved the problem in following steps:
1. Check if `chromedriver` installed correctly: these is no error when `npm install` and `PROJECT_ROOT\node_modules\chromedriver\lib\chromedriver\chromedriver.exe` exists.
2. Check if PATH wrong: `npm run demo` succeeded if adding `PROJECT_ROOT\node_modules\chromedriver\lib\chromedriver` dir to PATH env and failed if removing `PROJECT_ROOT\node_modules\chromedriver\lib\chromedriver` dir from PATH env.

So i am sure the problem is related to `PATH` im my circumstance. After some search, I found that https://www.npmjs.com/package/chromedriver#running-with-selenium-webdriver.
When `require('chromedriver')`, it will do https://github.com/giggio/node-chromedriver/blob/master/lib/chromedriver.js#L2. So `PATH` problem solved.

### DIFF
--- a/src/puppet-web/browser-driver.ts
+++ b/src/puppet-web/browser-driver.ts
@@ -81,6 +81,8 @@ export class BrowserDriver {
     }
    */
 
+    require('chromedriver')
+
     const options = {
       args: [
         '--homepage=about:blank',
--- a/src/puppet-web/browser-driver.ts
+++ b/src/puppet-web/browser-driver.ts
@@ -81,14 +81,6 @@ export class BrowserDriver {
     }
    */
 
-    /**
-     * https://github.com/Chatie/wechaty/pull/416
-     * In some circumstances, ChromeDriver could not be found on the current PATH.
-     * The chromedriver package always adds directory of chrome driver binary to PATH.
-     * So we requires chromedriver here to avoid PATH issue.
-     */
-    require('chromedriver')
-
     const options = {
       args: [
         '--homepage=about:blank',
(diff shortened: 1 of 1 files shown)


src/puppet-web/browser-driver.ts
```diff
@@ -81,6 +81,8 @@ export class BrowserDriver {
     }
    */
 
+    require('chromedriver')
```
zixia: Could you please add comments above this line to memo the details?

Becasue it looks wired if without any description.

Anyway, good catch and nice patch. Thank you!

xjchengo:
Comments have been added. But the test failed. I will try to figure it out later.

zixia:
Please let me know when you are ready.

xjchengo:
`wechaty` installs ChromeDriver using `apk add chromedriver` instead of `npm install chromedriver` in Docker.
So I don't require chromedriver package in Docker.
However, this time all failed tests threw:
```
✖ Exited because no new tests completed within the last 180000ms of inactivity
```
I don't know why they were timeout.

zixia:
1. The reason that we did not use `npm install chromedriver` is because npm chromedriver does not support alpine Linux. 

    I had submitted an issue before at here: https://github.com/giggio/node-chromedriver/issues/70

2. CI failed because there might have a bug in our code which will cause lost control of our code. 

    We have issue #175 for monitoring this.

3. Your PR looks good: it passed 3 tests:
    1. 1 on Appveyor: node@windows
    1. 2 on Travis CI: node v6@linux & v6@mac. (failed 2: node v7@linux & v7@mac)

So don't worry about the failed CI.

Before the PR can be merged, we need at least 3 approvements from @Chatie/contributor. I had already approved, so you need 2 more. After that, we can do merge. :) 

Thank you very much!

zixia:
@Chatie/contributor anyone who still using windows could check and approve this PR?

PR #440, opened 2017-04-26
add magic code for room.say()  when `@bot ` happen

lijiarui:
#362 
add m.mentioned() for message, thanks for @JasLin tell us about magic code when @ somebody,
so add maigic code in room.say()

### DIFF
--- a/src/room.ts
+++ b/src/room.ts
@@ -184,7 +184,8 @@ export class Room extends EventEmitter implements Sayable {
       const replyToList: Contact[] = [].concat(replyTo as any || [])
 
       if (replyToList.length > 0) {
-        const mentionList = replyToList.map(c => '@' + c.name()).join(' ')
+        const AT_SEPRATOR = String.fromCharCode(8197)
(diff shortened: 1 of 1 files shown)


src/room.ts
```diff
@@ -184,7 +184,8 @@ export class Room extends EventEmitter implements Sayable {
       const replyToList: Contact[] = [].concat(replyTo as any || [])
 
       if (replyToList.length > 0) {
-        const mentionList = replyToList.map(c => '@' + c.name()).join(' ')
+        const AT_SEPRATOR = String.fromCharCode(8197)
+        const mentionList = replyToList.map(c => '@' + c.name()).join(AT_SEPRATOR)
```
lijiarui: no, we don't need, `AT_SEPRATOR` equals space.

> wechat use special char with unicode code 8197 to indetify real@ 
by @JasLin

zixia:
Ok, ic. Please get at least one approvement before I could merge this PR.

Thanks!

zixia:
Thanks all for approving!

src/room.ts
zixia: OK

Issue #528, opened 2017-05-20
When a room member change it's nickname in a room, bot cannot detect it.


lijiarui:
When a contact change it's nickname in a room, using room.member() cannot get the contact successfully.

I guess this is because the following code in `room.ts` doesn't refresh contact
```ts
  private async readyAllMembers(memberList: RoomRawMember[]): Promise<void> {
    for (const member of memberList) {
      const contact = Contact.load(member.UserName)
      await contact.ready()
    }
    return
  }
```

zixia:
Yes, you are right.

lijiarui:
Maybe we should plan to do `refresh` in period inside the Wechaty Framework. Because all of room info should update.

It can add an enhancement.

zixia:
I will not disagree with you because I'd also like to consider adding a `refresh` in a period.

However, it will need not only a proposal in detail to discuss, but also more design patterns should be considered. I'd like to leave those works as an exercise to the community.

zixia:
Currently, we have to use `room.sync()` as a workaround. 

If anyone has better ideas, please let me know by leave comments under this issue, and I'll reopen it for discussion.

PR #529, opened 2017-05-20
add contact refresh when ready room member

lijiarui:
#528


zixia:
This is a very dirty fix.

It might work for you, but its pattern is not right.

I suggest you call `refresh` manually before there's other good way to do this.

PR #578, opened 2017-06-14
add notice to readme

imerse:


zixia:
This notice could be very useful for helping others to save time. 

Thank you very much!

PR #604, opened 2017-06-26
change outdated qrcode

lijiarui:
the previous QR code has more than 4700 friends and its login has been blocked by wechat, so change a new one here


zixia:
Thanks!

PR #627, opened 2017-07-01
fix #623

### LINKED ISSUE #623: [test] Unit Test for `mentioned` feature does not run at all
Get rid of all `setTimeout` to fix it.

https://github.com/Chatie/wechaty/blob/master/src/message.spec.ts#L226

```shell
  ✖ src › message › mentioned() Test finished without running any assertions
```

lijiarui:
fix #623 
delete setTimeout
modify mock_room error


zixia:
Looks good.

Could explain the reason of the failure before?

lijiarui:
mock room_data shouldn't be room.rawObj,  we should mock data structure as room.spec.ts do, but actual data should use room.rawObj.

I'm not sure where  `room.spec.ts`'s mock data structure from.

Issue #252, opened 2017-02-14
[doc] Contribute to the doc editing[ jsdoc / jsdoc2md / typedoc ]


ax4:
Hi, 

I find the API doc need:  
* some minor fix (such as the different naming [Message Class](https://github.com/wechaty/wechaty/wiki/API#message-class) & [Class Room](https://github.com/wechaty/wechaty/wiki/API#class-room)) 
* improvement on the formatting, e.g. the level setting of each title 
* maybe, add an index for better guiding

I am volunteering to contribute to doc. I was wondering what should I do?  Thanks!

zixia:
Hi @ax4 ,

Welcome!

An useful FrameWork with good document will help the community save lots of times. Wechaty is an useful ChatBot FrameWork, it need to be well ducumented.

We have a plan to re-document Wechaty, you can join us. Please see #73 .

Thank you very much for your volunteering!

lijiarui:
@ax4  We are trying to just give  outlined api in API doc, could you give an outline suggestion here?

ax4:
@lijiarui Hi, good to hear that!! I think, a good API doc should be based on User's most demand.  So I would suggest the outline as below:  

* Introduction to Wechaty
    * HelloWorld
    * Save my login status (profile)
* Basic Events of Wechaty
    * scan
    * login
    * logout
    * message
    * error
    * friend
    * room-join
    * room-leave
    * room-topic
* Reply a Message to Contact / Room 
    * Reply to Contact
    * Reply to Room
* Search for Contact / Room (Query Type)
    * Contact Query Type
    * Room Query Type
* API Details of Class
    * Wechaty Class
        * Wechaty.instance(profile:string): Wechaty
        * Wechaty.init(): Wechaty
        * Wechaty.send(message: Message): Wechaty
        * Wechaty.say(content: string)
    * Message Class
        * Message.from(contact?: Contact|string): Contact
        * Message.to(contact?: Contact|string): Contact | void
        * Message.content(content?: string): string
        * Message.room(room?: Room|string): Room | void
        * Message.type(): MsgType
        * Message.typeApp(): AppMsgType
        * Message.say(content: string): Promise
        * Message.self(message: Message): boolean
        * Message.id: string
        * Message.filename():string
    * Contact Class
        * Contact.id: string
        * Contact.name(): string | ''
        * Contact.city(): string | undefined
        * Contact.province(): string | undefined
        * Contact.gender(): Gender | undefine
        * Contact.avatar():
[...]

Issue #73, opened 2016-11-04
[doc] To Embed Document in Wechaty Code for Generating Automaticly


zixia:
The Public API of Wechaty is almost not changed much during the past five months, but the module is increasing, and the internal API is changed a lot.

We need to document them better, not only the public but also the internal modules and methods. 

So a document generator with embed document in code needs to be considering from now.

I'd like to prefer [Dgeni](https://github.com/angular/dgeni), which is the Documentation Generator of AngularJS/Angular Material/Protractor/Ionic.

~~Here're some useful links:~~
* ~~<https://github.com/angular/angular.js/wiki/Writing-AngularJS-Documentation>~~
* ~~<http://allenhwkim.tumblr.com/post/92161523693/sigh-angularjs-documentation>~~
---
### Update on Feb 2017:

Finally we decided to use the classic [jsdoc](http://usejsdoc.org/) to embed document in TypeScript, then generate document to [docs/](https://github.com/wechaty/wechaty/tree/master/docs) by [jsdoc2md](https://github.com/jsdoc2md/jsdoc-to-markdown).

* [An introduction to JSDoc](http://www.2ality.com/2011/08/jsdoc-intro.html)
* [Creates markdown API documentation from jsdoc-commented javascript](https://github.com/jsdoc2md/jsdoc-to-markdown)
* [Typescript: JSDoc support in JavaScript](https://github.com/Microsoft/TypeScript/wiki/JSDoc-support-in-JavaScript)

sleedata:
@zixia, I'll look into this.

kis87988:
I will try to contribute about this issue

zixia:
@kis87988 Hope you could help the community to get a better version of Documentations!

zixia:
Close this issue because the PR will be merged soon.

PR #640, opened 2017-07-08
add documentation TODO entries

hczhcz:
It would be helpful when using some documentation generators. At least, there would be a complete list of functions/methods

BTW, I would suggest that merging the [Wiki](/Chatie/wechaty/wiki/API) and the [`/docs`](/Chatie/wechaty/blob/master/docs/index.md) is good to new users.

### DIFF
--- a/src/contact.ts
+++ b/src/contact.ts
@@ -104,6 +104,9 @@ export class Contact implements Sayable {
   private dirtyObj: ContactObj | null
   private rawObj: ContactRawObj
 
+  /**
+   * @private
+   */
   constructor(
     public readonly id: string,
   ) {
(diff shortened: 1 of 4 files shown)


zixia:
Yes, I agree with you. The document should be generated from the source code, which will be fully synced with the latest version. We had some discussion about it, see #73 #252.

And I also agree with you about merging the [Wiki API](https://github.com/Chatie/wechaty/wiki/API) page to /docs. Because the Wiki page is created before we decide to use jsdoc to document Wechaty. Now we have the plan to do this, so I believe as soon as we merged the API docs to the jsdoc, we should redirect the Wiki page to /docs, which is good for new users.

I appreciate your PR for moving this forward: add a `@todo` is a good start. More contributions are very welcome. Thanks!

PR #651, opened 2017-07-12
room-join cannot detect inviteeList when people join in the room with qrcode

lijiarui:
room-join cannot detect inviteeList when people join in the room with QR code:

* If one contact joins in the room by the QR code shared by the bot, system message as follows:
   - "李佳芮" joined the group chat via your shared QR Code.
   - "李佳芮"通过扫描你分享的二维码加入群聊

* If one contact joins in the room isn't by the QRcode shared by the bot, system message as follows:
    - " 李佳芮" joined the group chat via your shared QR Code.
    - " 李佳芮"通过扫描你分享的二维码加入群聊
    (Pay attention that it has more space before 李佳芮)

### DIFF
--- a/src/puppet-web/firer.ts
+++ b/src/puppet-web/firer.ts
@@ -60,6 +60,7 @@ const regexConfig = {
 
   roomJoinQrcode: [
     /^" (.+)" joined the group chat via the QR Code shared by "?(.+?)".$/,
+    /^"(.+)" joined the group chat via the QR Code shared by "?(.+?)".$/,
     /^"(.+)" joined the group chat via "?(.+?)"? shared QR Code.$/,
     /^" (.+)"通过扫描"?(.+?)"?分享的二维码加入群聊$/,
     /^"(.+)"通过扫描"?(.+?)"?分享的二维码加入群聊$/,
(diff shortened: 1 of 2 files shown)


src/puppet-web/firer.ts
```diff
@@ -59,9 +59,10 @@ const regexConfig = {
   ],
 
   roomJoinQrcode: [
-    /^"(.+)" joined the group chat via the QR Code shared by "?(.+?)".$/,
-    /^"(.+)" joined the group chat via "?(.+?)"? shared QR Code.$/,
```
zixia: Why you get rid of `"?(.+?)"?`, and use `undefined` in your code, please explain.

I don't think this change is needed, and also I believe it's not a good choice to do this.

src/puppet-web/firer.ts
```diff
@@ -194,7 +195,9 @@ async function checkRoomJoin(m: Message): Promise<void> {
   let inviteeContactList: Contact[] = []
 
   try {
-    if (inviter === "You've" || inviter === '你' || inviter === 'your') {
```
zixia: I'd like to keep this line of code, instead of your new `!inviter`. 

Because I think to check more words in the message is a good way to make sure you are tracing the right message in Wechat.

zixia:
Thanks

PR #770, opened 2017-08-30
Fixed: replace `document.domain`  with `location.hostname` since front one sometimes lose subdomain name

zhenyong:


zixia:
Thanks for catching that!

I searched and find that someone said:
> Do not use window.document.domain as it is not supported consistently.
at https://github.com/zeroclipboard/zeroclipboard/issues/535

It seems we should switch to use `location.hostname` as you purposed.

Issue #70, opened 2016-11-03
Cannot create an instance of the abstract class 'FriendRequest


JasLin:
i am trying to test  FriendRequest in ``` example/ding-dong-bot.js`` , error occurs like that:

```
example/ding-dong-bot.ts (49,19): Cannot create an instance of the abstract class 'FriendRequest'. (2511)
```

code: 
```example/ding-dong-bot.js``` : 

```javascript
/**
 *
 * Wechaty - Wechat for Bot
 *
 * Connecting ChatBots
 * https://github.com/wechaty/wechaty
 *
 */

/* tslint:disable:variable-name */
const QrcodeTerminal = require('qrcode-terminal')

import {
  Wechaty
  , Config
  , log
  , FriendRequest
} from '../'

const welcome = `
| __        __        _           _
| \\ \\      / /__  ___| |__   __ _| |_ _   _
|  \\ \\ /\\ / / _ \\/ __| '_ \\ / _\` | __| | | |
|   \\ V  V /  __/ (__| | | | (_| | |_| |_| |
|    \\_/\\_/ \\___|\\___|_| |_|\\__,_|\\__|\\__, |
|                                     |___/

=============== Powered by Wechaty ===============
-------- https://github.com/zixia/wechaty --------

I'm a bot, my super power is talk in Wechat.

If you send me a 'ding', I will reply you a 'dong'!
__________________________________________________

Hope you like it, and you are very welcome to
upgrade me for more super powers!

Please wait... I'm trying to login in...

`

console.log(welcome)
const bot = Wechaty.instance({ profile: Config.DEFAULT_PROFILE })

bot
.on('login'	  , user => {
  log.info('Bot', `${user.name()} logined`)
  const request = new FriendRequest()
  console.log('loging....................................')
  console.log(request)
})
.on('logout'	, user => log.info('Bot', `${user.name()} logouted`))
.on('error'   , e => log.info('Bot', 'error: %s', e))
.on('scan', (url, code) => {
  if (!/201|200/.test(String(code))) {
    let loginUrl = url.replace(/\/qrcode\//, '/l/')
    QrcodeTerminal.generate(loginUrl)
  }
  console.log(`${url}\n[${code}] Scan QR Code in above url to login: `)
})
.on('message', m => {
  try {
    const room = m.room()
    console.log((room ? '[' + room.topic() + ']' : '')
                + '<' + m.from().name() + '>'
                + ':' + m.toStringDigest()
    )

[...]

PR #812, opened 2017-09-17
Friendrequest doc

lijiarui:
add jsdoc for friend request


zixia:
The `FriendRequest` design has some problems which is not resolved yet. 

See: #70 

However, this is a good workaround for the document, so let's merge it first.

Thanks!

PR #823, opened 2017-09-20
Update readme API automatically

lijiarui:
### commit description
* In order to sync `README.md` API doc with code,  I create a README template in `docs/partials/REAMDE.hbs`, see more in [jsdoc-to-markdown-wiki-create-readme-template](https://github.com/jsdoc2md/jsdoc-to-markdown/wiki/Create-a-README-template)

* Then I got a problem: we cannot sync all of wechaty API in README file because it is so big.    
So I picked all of class & function name in README and their links.

* And then I got another problem: the function link generated by jsdoc automatically is anchor link (`#`),  it means I cannot link it to other pages(http://chatie.io/wechaty), this is very inconvenient for  readers, so I override the following templates in `docs/partials/overrides`, adding http://chatie.io/wechaty in the link:
    * link.hbs
    * sig-link-parent.hbs

### scripts in package.json description

It will generate two docs:

* `npm run readme-doc`  will generate README.md using `README.hbs` template and using the override file `link.hbs` & `sig-link-parent.hbs` to add `http://chatie.io/wechaty` in the doc link.
```
 "readme-doc": "jsdoc2md --partial docs/partials/overrides/*.hbs --template docs/partials/README.hbs dist/src/{wechaty,room,contact,friend-request,message}.js dist/src/puppet-web/friend-request.js>> README.md",
```

* `npm run api-doc` will generate `index.md` as we used before.
```
jsdoc2md dist/src/{wechaty,room,contact,friend-request,message}.js dist/src/puppet-web/friend-request.js>> docs/index.md
```


zixia:
The new api index document seems great.

However I feel this template  is some how too complicated for the readme, will think about it later.

lijiarui:
Trying to make it simpler:

* run `npm run doc` can generate `REAMDE.md` and `index.md`
* move `REAMDE.hbs` to root directory

If user want to modify README file，one  just modify the root directory file `README.hbs` and run `npm run doc`, `README.hbs` also the markdown format.

zixia:
Thank you for trying to make it simple.

It's good to embed the jsdoc TOC to the readme. But what I want to do is to _keep the README.md as it is_. 

Could we update README.md by an npm script, instead of totally replace it by a template?

docs/partials/link.hbs
```diff
@@ -0,0 +1,25 @@
+{{! usage: link to="namepath" html=true/false caption="optional caption"~}}
+
+{{~#if html~}}
+<code>
+
+{{~#link to~}}
+{{#if url~}}
+<a href="http://www.baidu.com/{{{url}}}">{{#if ../../caption}}{{../../../caption}}{{else}}{{name}}{{/if}}</a>
```
zixia: baidu? WTF?

package.json
```diff
@@ -109,9 +113,10 @@
     "@types/node": "8.0.28",
     "bl": "1.2.1",
     "body-parser": "1.18.0",
-    "chromedriver": "2.32.3",
     "brolog": "1.2.6",
+    "chromedriver": "2.32.3",
     "express": "4.15.2",
+    "markdown-include": "^0.4.3",
```
zixia: I'd like to do not add ` markdown-include` dependence because it seems the job could be done without it.

markdown.json
```diff
@@ -0,0 +1,7 @@
+{
+	"build" : "README.md",
+	"files" : ["README.md"],
+	"tableOfContents": {
```
zixia: Could we do the job without this file?

lijiarui:
Thanks to @binsee 's idea.

I add markdown comment in README.md as follows:
```
[comment]: # (JSDOC SYNC BEGIN) 
[comment]: # (JSDOC SYNC END) 
```
and change the code between the comment automatically using script.
For developer, if the want sync api, just run `npm run doc` is ok, it can sync both full api and readme api index.

About the script description:
* npm run doc = npm run doc-api + npm run doc-readme   
* npm run doc-api: generate full api doc.   
* npm run doc-readme = npm run doc-index + ts-node docs/update-readme.ts + npm run update-readme + remove doc-index.md
* npm run doc-index: generate doc index for readme file
* ts-node docs/update-readme: get the code between the comment and add the doc index generated automatically.
*  npm run update-readme: use [markdown-include](https://github.com/sethen/markdown-include) to add api-index-doc into readme

docs/partials/link.hbs
lijiarui: o, it's just a test file... I remembered I deleted...

package.json
lijiarui: nop, this dependence is necessary. Or I have to write something like this.

package.json
```diff
@@ -18,7 +18,10 @@
     "ava": "ava --verbose --extension ts",
     "ts-node": "ts-node",
     "dist": "npm run clean && tsc && jq \"del (.files)\" < package.json > dist/package.json && shx cp src/puppet-web/*.js dist/src/puppet-web/",
-    "doc": "npm run dist && echo '# Wechaty v'$(jq -r .version package.json)' Documentation\n' > docs/index.md && jsdoc2md dist/src/{wechaty,room,contact,friend-request,message}.js dist/src/puppet-web/friend-request.js>> docs/index.md",
+    "doc": "npm run dist && echo '# Wechaty v'$(jq -r .version package.json)' Documentation\n' > docs/index.md && npm run doc-api && npm run doc-readme",
```
binsee: run `npm run doc`, output:

```cmd
E:\CodeWork\fork\wechaty>npm run doc

> wechaty@0.8.231 doc E:\CodeWork\fork\wechaty
> npm run dist && echo '# Wechaty v'$(jq -r .version package.json)' Documentation
' > docs/index.md && npm run doc-api && npm run doc-readme


> wechaty@0.8.231 dist E:\CodeWork\fork\wechaty
> npm run clean && tsc && jq "del (.files)" < package.json > dist/package.json && shx cp src/puppet-web/*.js dist/src/puppet-web/


> wechaty@0.8.231 clean E:\CodeWork\fork\wechaty
> shx rm -fr dist/*

'# Wechaty v'$(jq -r .version package.json)' Documentation
```

Under Windows, \n will parse the error, causing the execution to be interrupted

In addition, echo command in the windows and linux system implementation of the results are different 😂 

```cmd
E:\CodeWork\fork\wechaty>echo '# Wechaty v'$(jq -r .version package.json)' Documentation'
'# Wechaty v'$(jq -r .version package.json)' Documentation'

E:\CodeWork\fork\wechaty>bash
li@BinWork:/mnt/e/CodeWork/fork/wechaty$ echo '# Wechaty v'$(jq -r .version package.json)' Documentation'
# Wechaty v0.8.231 Documentation
```

binsee: System differences, there is no way

Generating readme very well

zixia: `npm run doc:api`
and
`npm run doc:readme`

package.json
```diff
@@ -18,7 +18,10 @@
...
+    "doc-readme": "npm run doc-index && ts-node docs/update-readme.ts && shx rm -rf docs/doc-index.md",
+    "doc-index": "jsdoc2md --partial docs/partials/*.hbs --template docs/api-index.hbs dist/src/{wechaty,room,contact,friend-request,message}.js dist/src/puppet-web/friend-request.js>> docs/doc-index.md",
```
zixia: Please:
1. rename `doc-index`  to `doc:toc`
1. output `doc:toc` to `stdout` instead of writing it to file, because its output is for temporary usage.

docs/update-readme.ts
```diff
@@ -0,0 +1,12 @@
+const fs = require('fs')
+async function updateReadme() {
+  const readme = await fs.readFileSync('README.md')
+  const beginIndex = readme.toString().search(/\[comment\]\: # \(JSDOC SYNC BEGIN\)/)
+  const endIndex = readme.toString().search(/\[comment\]\: # \(JSDOC SYNC END\)/)
+  const docString = readme.toString().substring(beginIndex, endIndex)
+
+  const docIndex = await fs.readFileSync('docs/doc-index.md')
```
zixia: Please use `stdin` to read the TOC content at here.

docs/api-index.hbs
```diff
@@ -0,0 +1,25 @@
+{{#class name="Wechaty"~}}
```
zixia: Please move all `*.hbs` file to `docs/config/` directory, to give them a more clear location to self-document.

docs/partials/sig-link-parent.hbs
```diff
@@ -0,0 +1,13 @@
+{{#if name}}{{#sig~}}
```
zixia: What's this file for? Please explain it(and better to document it).

docs/update-readme.ts
```diff
@@ -0,0 +1,12 @@
...
+  const insertToReadme = readme.toString().replace(docString, '[comment]: # (JSDOC SYNC BEGIN) \n\n' + docIndex + '\n')
+  await fs.writeFileSync('README.md', insertToReadme, 'utf8')
+}
+updateReadme()
```
zixia: Please move this file to `/script/` because it should not under the docs directory.

Or to make it better: could you get rid of this file, and do this job by npm script? If you could, that would be great.

docs/partials/sig-link-parent.hbs
lijiarui: This file is the jsdoc2md official partials: https://github.com/jsdoc2md/dmd/blob/master/partials/shared/signature/sig-link-parent.hbs

I  add link `chatie.io/wechaty` here to custom the linkable toc

docs/api-index.hbs
lijiarui: And this `hbs` is not official part. So I suggest to seperate `api-index.hbs` with the other two

lijiarui:
About this commit
### using gsed for GNU version of sed under mac
Under  MacOs using `gsed` to enhance sed :  see [stackoverflow-sed-command-with-i-option-failing-on-mac-but-works-on-linux](https://stackoverflow.com/questions/4247068/sed-command-with-i-option-failing-on-mac-but-works-on-linux)

btw, because not that familiar to shell, I thought it should integrate to one command using gsed, but I failed to do than and it keep on several command...

### should seperate the two kind of `.hbs` file
There are two kind of partial:

* One is the rewrite system template----under config file: `link.hbs` and `sig-link-parent.hbs`

* One is the template I wrote for doc-toc: doc-toc.hbs

So I seperate the two hbs file

zixia:
I still feel there's something not very right in this PR:
1. It's better not to use any additional package.
1. It's better to not to rewrite system template. The benefit is that we will not add those two new files.
1. It's better to use pipe instead of a temporary file.

However, I'll merge this PR if you could get three approvements from the contributors.

Good job, thanks.

lijiarui:
will use this https://github.com/jonschlinkert/markdown-toc in the future.

It seems beautiful in my repo: https://github.com/lijiarui/tencent-nlp

PR #1014, opened 2017-12-15
Update tuling123-bot.ts

htoooth:
fix tuling123-bot error

### DIFF
--- a/example/tuling123-bot.ts
+++ b/example/tuling123-bot.ts
@@ -82,7 +82,7 @@ bot
   log.info('Bot', 'talk: %s'  , msg)
 
   try {
-    const reply = tuling.ask(msg.content(), {userid: msg.from()})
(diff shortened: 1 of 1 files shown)


zixia:
Thanks for the PR.

Does this mean the tuling API client changed its behaviour?

htoooth:
I found that tuling123'api is not changed.  This example always run error.

I check this module. Tuling123'api `ask` always return `Promise`. And Now,  the tuling bot's result is `{code: number, text: string}`. you should have a try. 

Test case is not pass. I don't know why. I'm not change other file.  This testcase is not pass in 
`test/fixture/docker/type-error.ts` file: 

```js
let i = 3
i = "should set to string because it's type is number"
console.log(i)
```

zixia:
@htoooth Sorry for the late reply.

I'll merge this and have a try.

Thank you very much for the contribution!

Issue #594, opened 2017-06-20
[enhancement] Add timezone in to Dockerfile


kis87988:
Right now, we can not declare a environment (ENV) varialbe to set timezone
My idea is:
> 1. apk add tzdata
> 2. set ENV localTime America/Los_Angeles
> 3. RUN cp /usr/share/zoneinfo/${localTime} /etc/localtime
but I could not figure it out how to make it run in the dockerfile

The alternating method is:
create container first by docker pull/docker run and then docker enter to the container.
do the following: 
> 1. apk add -U tzdata
> 2. cp /usr/share/zoneinfo/America/Los_Angeles /etc/localtime

Hope someone can make it into dockerfile.

zixia:
Thanks for purposing!

I believe setting timezone is a necessary feature that Wechaty Docker Image should support.

zixia:
Support to set `TZ` for timezone at Wechaty version 0.13.13 or above at: https://github.com/Chatie/wechaty/commit/61c277c4b3400c155544fec59f8e2659238352f4

Usage:

```shell
docker run -e TZ='Asia/Shanghai' zixia/wechaty
```

PR #1028, opened 2017-12-21
added TZ to set timezone for container

Gcaufy:
```
$ docker run -d --volume="$(pwd)":/bot -e TZ=Asia/Shanghai zixia/wechaty index.js
```

The container default timezone is "Etc/UTC". when I was using wechaty to do some time base scheduled works I have to calculate the timezone differences.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,6 +2,7 @@ FROM ubuntu:17.10
 LABEL maintainer="Huan LI <zixia@zixia.net>"
 
 ENV DEBIAN_FRONTEND     noninteractive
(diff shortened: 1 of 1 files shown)


zixia:
We had already supported TZ(TimeZone) by #594 , Had you tried that?

If after you tested that and you believe we still need this PR, please let me know. 

Thanks!

Gcaufy:
Confirmed, thx

zixia:
Glad to hear that. Thanks for your PR anyway!

I believe we need another PR top document how to use `TZ` for TimeZone settings. @lijiarui

PR #1087, opened 2018-02-01
Add troubleshooting for hidden .txt files (Windows)

IdiosApps:
Could help users start making chatbots even faster!


zixia:
Thanks for trying to make Wechaty to be better!

PR #1119, opened 2018-03-07
add one more test case for digestEmoji(), close #1108

### LINKED ISSUE #1108: emoji suggestion
I found [src/misc](https://github.com/Chatie/wechaty/blob/master/src/misc.ts) is trying to strip emoji, HTML and any misc things in a text message. 

I have 2 suggestions:

1.  I feel a bit confused about all emoji rules and this file rules, so I suggest to add some comment here, then if any other one gets some new emoji rules, we can add it here.

2. I suggest to add two functions:
* `message.stripEmoji(): void`: to strip emoji from the text message, then we can get a pure one.
* `message.emoji(): []`: return a list containing all emojis meaning in the text.

I found the following code in [src/misc](https://github.com/Chatie/wechaty/blob/master/src/misc.ts) seems trying to get emoji meaning but it cannot always succeed, also if there is some emotion, like gif in wechat, it doesn't work too. 
```ts
  public static digestEmoji(html?: string): string {
    if (!html) {
      return ''
    }
    return html
          .replace(/<img class="(\w*?emoji) (\w*?emoji[^"]+?)" text="(.*?)_web" src=[^>]+>/g,
                   '$3',
                 ) // <img class="emoji emoji1f4a4" text="[流汗]_web" src="/zh_CN/htmledition/v2/images/spacer.gif" />
          .replace(/<span class="(\w*?emoji) (\w*?emoji[^"]+?)"><\/span>/g,
                   '[$2]',
                 ) // '<span class="emoji emoji1f334"></span>'
  }
```

Well,  thinks more, even it should be a seperate npm module....

suntong:


suntong:
@zixia, I've added one more test case for digestEmoji():

    

let's close #1108 for now, until something concrete comes up.

suntong:
hold on, PR contaminated...

suntong:
>
> hold on, PR contaminated...
>

@zixia, OK I send the PR via email instead?

The PR from github should only pull in the `emoji` branch, but it always
include the changes I made in other branches as well. So sending the PR via
email is the only way. thx.

zixia:
No, PR via email should not be accepted because that will leak much important information for the code.

I'd like to suggest you close this PR 1st, then create a new branch from the latest version of master, to send your PR again.

suntong:
k. will do.

PR #1129, opened 2018-03-16
- [+] add one more test case for digestEmoji(), close #1108

### LINKED ISSUE #1108: emoji suggestion

suntong:


zixia:
Thank you!

PR #1133, opened 2018-03-17
fixed. user undefined on example contact-bot

Lupino:


zixia:
Thanks for the PR!

I remembered that in TypeScript there's a kind of style define the type of `this` as the following code demostrated:

https://github.com/Chatie/wechaty/blob/34efb96dc0a322793325438d78e9c8ad7fae8f38/src/wechaty.ts#L235-L246

Could you please provide how to reproduce `user undefined` error to make it clear?

zixia:
Hello @Lupino,

Recently we had moved all the examples to the starter repo at <https://github.com/Chatie/wechaty-getting-started/tree/master/examples>.

Please feel free to try the examples over there, and let me know if you have any problem with them.

About this PR, I had confirmed that the `function(this: Wechaty, ...)` is a syntax sugar of TypeScript, so it should work without any problem.

BTW: I had saw that you build a Wechat CLI tool based on Wechaty, it's very interesting. I had add it to our Wiki at <https://github.com/Chatie/wechaty/wiki/PoweredByWechaty> and please feel free to modify it as you like.

PR #1341, opened 2018-06-14
Bug compatible WXCreateChatRoom

lijiarui:
see https://github.com/lijiarui/wechaty-puppet-padchat/issues/62


zixia:
Please test your code and confirm it works as you expected.

I do not believe it does what you want it to.

zixia:
This PR looks better after you commit your new modifications.

I'd like to suggest that we should write a new pure function helper to implement this function, maybe `fixWei()` in `fix-wei.ts` is a good name for it, but it's just a suggestion.

And we need to bundle unit tests with the new function, I think there's already two in your comments:

```
   // BUG compitable: "\n\u00135907139882@chatroom" -> "5907139882@chatroom"
   // BUG compitable: "\n\u001412558026334@chatroom" -> "12558026334@chatroom"
   str.replace(/^\n[\u0000-\uffff]/g, '')
```

lijiarui:
Good suggestion! I suggest:
filename: `compatible-wei.ts`
function: `pureuserName()`

zixia:
Ok. 

I'll merge this PR after you finished it.

src/puppet-padchat/pure-function-helpers/index.ts
```diff
@@ -24,3 +24,4 @@ export * from './room-event-leave-message-parser'
 export * from './room-event-topic-message-parser'
 export * from './room-raw-payload-parser'
 export * from './split-name'
+export * from './compatible-wei'
```
zixia: Please modify the export files list in order: by filename.

src/puppet-padchat/pure-function-helpers/compatible-wei.ts
```diff
@@ -0,0 +1,20 @@
...
+ * WXCreateChatRoom() will get dirty chatroom Id.
+ * function pureUserName can get a clean chatroom Id.
+ * https://github.com/lijiarui/wechaty-puppet-padchat/issues/62
+ *
+ * WXCreateChatRoom result:
+ * {"message":"\n\u0010Everything is OK","status":0,"user_name":"\n\u00135907139882@chatroom"}
+ * BUG compitable: "\n\u00135907139882@chatroom" -> "5907139882@chatroom"
+ * BUG compitable: "\n\u001412558026334@chatroom" -> "12558026334@chatroom"
+ */
+export function pureUserName(id?: string) {
```
zixia: pureUserName -> stripBugWei

lijiarui: Maybe not `stripBugWei `, because I'm sure it will have more bug in the future...

src/puppet-padchat/pure-function-helpers/index.ts
```diff
@@ -12,6 +12,7 @@
  *
  */
 
+export * from './compatible-wei'
```
zixia: compatible-wei -> compatible-wei-bug

zixia: This rename is required because that we compatible with wei because it has BUG.

lijiarui: you are right, done

Issue #1346, opened 2018-06-15
Prevent the Floating Promise in the Async/Await Code


zixia:
According to #1289, we should always await a Promise no matter than what because it will prevent we to make very hard to track mistakes.

So we should enable tslint to check the floating promise for us. So should all the developers do if you are writing async/await code:

Related configuration in ***tslint.json***:

```json
{
    "extends": "tslint:recommended",
    "rules": {
        "no-floating-promises": true
    }
}
```

### See Also

- [Detect missing await in TypeScript](https://www.meziantou.net/2017/11/20/detect-missing-await-in-typescript)

zixia:
@lijiarui Could you please help me to check out the latest master branch of code, and fix all the warnings when running `npm run lint`, then send me back a PR to fix them?

Thanks.

lijiarui:
see #1348

zixia:
Thank you very much.

PR #1348, opened 2018-06-15
fix warnings when run `npm run lint`

lijiarui:
#1346 

Besides the following warning:
```
ERROR: /Users/egg/git/rui/wechaty/scripts/sort-contributiveness.ts[109, 1]: Promises must be handled appropriately
ERROR: /Users/egg/git/rui/wechaty/scripts/update-license.ts[149, 1]: Promises must be handled appropriately
```

I don't know how to add `await` in the function like this:
```ts
async function main(): Promise<void> {
  const pattern = '{bin/**/*.ts,examples/**/*.{js,ts},scripts/**/*.{ts,js},src/**/*.{ts,js},tests/**/*.ts}'
  // const pattern = 't.ts'
  const srcFileList = await glob(pattern)
  const promiseList = srcFileList.map(updateLicense)
  await Promise.all(promiseList)
}

main()
```

### DIFF
--- a/src/puppet-padchat/padchat-manager.ts
+++ b/src/puppet-padchat/padchat-manager.ts
@@ -221,7 +221,7 @@ export class PadchatManager extends PadchatRpc {
 
     const succeed = await this.tryAutoLogin(this.memorySlot)
     if (!succeed) {
-      this.startCheckScan()
+      await this.startCheckScan()
     }
 
     this.state.on(true)
@@ -239,12 +239,12 @@ export class PadchatManager extends PadchatRpc {
       log.warn('PuppetPadchatManager', 'stop() subscript not exist')
     }
 
-    this.stopCheckScan()
+    await this.stopCheckScan()
 
     // await this.padchatRpc.stop()
     await super.stop()
 
-    this.releaseCache()
+    await this.releaseCache()
 
     this.selfId          = undefined
     this.loginScanQrcode = undefined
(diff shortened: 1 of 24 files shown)


zixia:
Good job.

I had just fixed the two files that you did not fix in this PR, please modify this PR and make sure the CI turns GREEN before we could be able to merge it.

Thank you very much.

src/puppet-padchat/padchat-manager.ts
```diff
@@ -221,7 +221,7 @@ export class PadchatManager extends PadchatRpc {
 
     const succeed = await this.tryAutoLogin(this.memorySlot)
     if (!succeed) {
-      this.startCheckScan()
+      await this.startCheckScan()
```
zixia: This might block the start() for a long time which you will not want to, please test it before send to PR.

I would like to suggest you to try using `this.startCheckScan().catch(e => log.warn('...'))` and it might be able to work.

lijiarui:
When I run `npm run lint`, I found a lot of error related to the following type in `src/puppet/schema.ts`:
```ts
export interface MessagePayloadBase {
  id            : string,
  contactId?    : string,        // Contact ShareCard
  filename?     : string,
  mentionIdList : string[],      // Mentioned Contacts' Ids
  text?         : string,
  timestamp     : number,        // Unix Timestamp(in seconds)
  type          : MessageType,
}
```

It is said `mentionIdList` cannot be undefined. Is this designed to do this or by mistake?

zixia:
It's designed and need you first the parser pure function to compatible with it.

Better to implement the actual function is you could, however you can also just put a empty array in it and we can implement it later.

Issue #1345, opened 2018-06-15
When bot quit the room, bot still thought it in the room.


lijiarui:
#### Wechaty version:
0.15.172

#### Bug
When bot quit the room, bot still thought it in the room.

#### Reproduce:

1. run `room-bot.ts`
2. send ding to the bot, the bot create dingRoom
3.bot quit dingRoom
4.send ding to the bot again

#### Expect
the bot will create a ding Room again

#### Actual
bot says it has already had ding room and the sender is already in dingRoom

#### Full log

```shell
11:03:29 VERB PuppetPadchat onPadchatMessage({id=38815767555888173, type=Text(5)})
rawPayload: { content: 'ding',
  continue: 1,
  description: 'wechaty-alias : ding',
  from_user: 'qq512436430',
  msg_id: '38815767555888173',
  msg_source: '<msgsource />\n',
  msg_type: 5,
  status: 1,
  sub_type: 1,
  timestamp: 1529031808,
  to_user: 'wxid_zj2cahpwzgie12',
  uin: 324216852 }
11:03:29 VERB Message static create(38815767555888173)
11:03:29 SILL Accessory #14<Message> constructor()
11:03:29 VERB Message constructor(38815767555888173) for class Message
11:03:29 VERB Message ready()
11:03:29 SILL Puppet messagePayloadCache(38815767555888173) cache MISS
11:03:29 VERB Puppet messagePayload(id=38815767555888173, dirty=false)
11:03:29 SILL Puppet messagePayloadCache(38815767555888173) cache MISS
11:03:29 VERB PuppetPadChat messageRawPayloadParser({msg_id="38815767555888173"})
11:03:29 SILL PuppetPadchat messagePayload({"id":"38815767555888173","timestamp":1529031808,"type":6,"fromId":"qq512436430","toId":"wxid_zj2cahpwzgie12","text":"ding"})
11:03:29 SILL Puppet messagePayload(38815767555888173) cache SET
11:03:29 SILL Contact ready() @ PuppetPadchat<default>
11:03:29 SILL Contact ready() isReady() true
11:03:29 SILL Contact ready() @ PuppetPadchat<default>
11:03:29 SILL Contact ready() isReady() true
11:03:29 SILL Contact alias()
11:03:29 SILL Contact alias()
<李佳芮>:Message#Text(Contact<wechaty-alias>►Contact<苏轼>)<ding>
11:03:29 VERB Puppet selfId()
[...]

PR #1352, opened 2018-06-15
check room valid by id

lijiarui:
#1345 

This is because when bot quit the room, room data still in cache. We should call `WXGetChatRoomMember(roomId: string)` to check whether chatroom still valid. See more data in : https://github.com/lijiarui/wechaty-puppet-padchat/issues/38


zixia:
Related with https://github.com/lijiarui/wechaty-puppet-padchat/issues/64

PR #1374, opened 2018-06-21
Fix room.add() failed when room member more than 40

lijiarui:
see https://github.com/lijiarui/wechaty-puppet-padchat/issues/70

  - When member more than 40, use WXInviteChatRoomMember
  - When member less than 40, use WXAddChatRoomMember

So add WXInviteChatRoomMember in RPC function.


src/puppet-padchat/padchat-rpc.ts
```diff
@@ -783,6 +783,16 @@ export class PadchatRpc extends EventEmitter {
       return -1
     }
 
+    /**
+     * see https://github.com/lijiarui/wechaty-puppet-padchat/issues/70
+     * If room member more than 40
+     * Need call `WXInviteChatRoomMember` instead `WXAddChatRoomMember`
+     */
+    if (result.status === -2012) {
```
zixia: no magic number please.

lijiarui:
WXRoomAddTypeStatus type, maybe need to fill more in the future.

```ts
export enum WXRoomAddTypeStatus {
  Done          = 0,
  NeedInvite    = -2012,
  InviteConfirm = -2028,
}
```

PR #1402, opened 2018-06-27
sync contact and room per hour

lijiarui:
r't


zixia:
Please pay attention that the sync task might not finish after 1 hour.

The better way to do this is to set a timer after the sync finish.

I'll merge this PR but please refactor it later.

PR #1498, opened 2018-07-23
add Contact self class

lijiarui:
see https://github.com/qhduan/wechaty-doc/issues/25


zixia:
Thanks for follow up with the latest code!

Issue #1500, opened 2018-07-24
BREAKING CHANGE: v0.20 `Contact.alias` changed from Sync to Async


zixia:
### Before

```ts
const alias = contact.alias()
```

### After

```ts
const alias = await contact.alias()
```

PR #1502, opened 2018-07-24
Contact get alias from sync to async

lijiarui:
https://github.com/Chatie/wechaty/issues/1500

@zixia maybe you should change jsdoc at the same time when you have break change next time ：）


zixia:
Thanks

PR #1514, opened 2018-07-30
remove useless param type in example code

lijiarui:


zixia:
Could you explain why you decide to remove all the typing annotations in the examples code?

lijiarui:
For a newbie, the will run the example code, and most of them use don't very familiar with ts, when they copy the full example and don't know how to settings, they will get something wrong like the following error:
5541532955122_ pic
5531532955071_ pic

So in order to let a newbie run faster, I think we should remove this. For the developer who is familiar to wechaty, maybe they don't need the example code and can use typescript know what param is what type.

zixia:
Thanks for the explanation, it's very reasonable.

Merged.

PR #1538, opened 2018-08-10
add receive and send link prototype

windmemory:
This is a prototype of dealing with link message. 

My proposal design here is adding another field in the message, which is `linkPayload`, and store the content inside it. And when user want to send link message, then can `say` the `linkPayload`, then `wechaty` will send a link message. 

link

The structure of the `linkPayload` is below, refer to the screenshot above
```typescript
export interface LinkPayload {
  des?: string,
  thumburl?: string
  title: string,
  url: string,
}
```

The reason I didn't combine other card with link is that I found different type of card has different properties, if we put them all together, and leave some property as optional, it might cause confusion, so I added the specific payload for link. 

In this change, I just implemented the `Message.say()`, since `Contact` or `Room`'s `say()` method are similar to each other. 

@zixia 
Please let me know your thought about this implementation. And I used Amazon cloud9 environment this time, here is the [link](https://us-west-2.console.aws.amazon.com/cloud9/ide/230d341d0a924ceda2f5d907e3bd4769)

I've sent out an invitation email to you, please use that credential to access the cloud9. Let me know if you can not access the cloud9.

### DIFF
--- a/src/user/message.ts
+++ b/src/user/message.ts
@@ -47,6 +47,7 @@ import {
 import {
   MessagePayload,
   MessageType,
+  LinkPayload,
 }                 from 'wechaty-puppet'
 
 /**
@@ -327,6 +328,7 @@ export class Message extends Accessory implements Sayable {
   public async say (text:    string, mention?: Contact | Contact[]) : Promise<void>
   public async say (contact: Contact)                               : Promise<void>
   public async say (file:    FileBox)                               : Promise<void>
+  public async say (link:    LinkPayload)                           : Promise<void>
 
   public async say (...args: never[]): Promise<never>
   /**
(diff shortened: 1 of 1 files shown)


zixia:
Thanks for the PR.

However, please move all the discussion to an issue before you create a PR, and only keep the issue number in the PR content, which I think it's better to show what's happened in the CHANGELOG.

windmemory:
Got it, will create issue next time.

windmemory:
It seems like the current design for `UrlLink` is that, once the link is created, the content(title, description, url, thumbnail url) can not be modified. Can we make the properties modifiable? I think people might want to have control on that.

zixia:
The best practice is to keep the date immutable as possible as we can.

If you want to change those values, you can do that by creating a new UrlLink.

Or could you provide a scenario that can prove we have to support mutation?

windmemory:
But even create a new UrlLink, the `title`, `thumbnailurl` and `description` is immutable.

I think either we support modifying the link or creating new `UrlLink` with full control.

I can not come up with use cases modifying the `UrlLink`, but create a customize link should be common scenario. 

What do you think?

zixia:
Please file another issue for this immutable discuss, let's track it more clear.

PR #1555, opened 2018-08-16
do roomMemberPayloadDirty when sync room

windmemory:
In order to solve issue: https://github.com/Chatie/wechaty-puppet/issues/23

This change rely on PR: https://github.com/Chatie/wechaty-puppet/pull/25

### DIFF
--- a/src/user/room.ts
+++ b/src/user/room.ts
@@ -328,6 +328,7 @@ export class Room extends Accessory implements Sayable {
 
     if (dirty) {
       await this.puppet.roomPayloadDirty(this.id)
+      await this.puppet.roomMemberPayloadDirty(this.id)
     }
     this.payload = await this.puppet.roomPayload(this.id)
 


windmemory:
AFAIK, there are some changes we don't get notifications, such as someone leave the room by themselves. So there will be no way for us to keep the cache, no matter which layer's cache, up to date. Unless we do such `sync` operations for each call. 

So I think leaving this option to our developers does make sense. If they want to refresh the data, they are able to do that. 

---
If you insist on controlling all the data refresh inside `wechaty`, then we need to discuss this into more details. Actually I've come along to this point early today.

Here are my thoughts:
* We need to think about all the operations into two groups
  * Operations only involve one record (contact or room)
  * Operations involve multiple records, or even all the records
* For the first kind of operations, we always get the latest data from remote
* For the second kind of operations, we use our cache to reduce the server load
* For the room data, we treat them into two layers, the room itself and the room member list data, just like the one we are doing now. But if the user ask for a single room's payload, we get the latest one and return back, besides, we also check the roomMemberList with the cached one, if they are different, we refresh the room member list at the same time. 
* When events are triggered, we updated corresponding data with the data from the event.

What do you think about this?

windmemory:
I think this PR is a short term solution, and for long term, we might want to rethink about the design of how we deal with the cache, which is a big change. Currently the room member information can not be updated from the developer's perspective unless they restart `wechaty`. So I think this short term fix can be merged with an `Temp Fix` remark on it.

zixia:
> So there will be no way for us to keep the cache, no matter which layer's cache, up to date.

Yes, I agree with you and this is the biggest chanllenge if we get rid of the `sync()` method. That's why it is still there, or I had already purged it long times ago.

> We need to think about all the operations into two groups: involve one vs multiple

According to your design, how could you identify which operation is one, and which is multiple, given the following code?

```ts
roomList = await wechaty.Room.findAll() // 1
roomList().forEach(room => {
  console.log(await room.topic())  // 2
  const memberList = await room.memberList() // 3
  memberList.forEach(member => console.log(member.name())) // 4
})
```

My feeling is it will be very hard for the framework to classify it properly.

zixia:
One more thing: We can never assume that the developers know what he is doing...

2/3/4 will cause huge problems for your server: thoughsands of the traffic will go directly to the server in concurrent.

For example, if a developer wants to output all the room topics that he has, and he has 1,000 rooms. According to the above code, his code will make Wechaty call your server roomPayload API 1,000 times in parallel.

Issue #1604, opened 2018-09-26
room mention contact should using roomAlias


kis87988:
Purpose:
For now, 
Our room.say(content,contact) will call contact name, but if contact has room alias, it should do room alias. 
the order should be roomAlias > name. If the contact not set roomAlias, it would be name.
Is this a good idea? if so, I can do this PR

zixia:
Yes, the idea is the right direction that we should go.

PR is welcome.

zixia:
PR merged.

Thanks for your contribution!

kis87988:
I see! That is why. 
Thank you. I have learned a lesson 👍

zixia:
We had added an additional check for this special case: we do not allow to use promise as boolean anymore.

```
ERROR: /Users/zixia/chatie/wechaty/src/user/room.ts[437, 52]: Promises should be awaited when used in boolean expressions
```

See: https://github.com/Microsoft/tslint-microsoft-contrib/issues/174#issuecomment-362366666

PR #1605, opened 2018-09-27
fix: https://github.com/Chatie/wechaty/issues/1604

kis87988:
## I'm submitting a...

```
[ ] Bug Fix
[x] Feature
```

## Checklist

  - A bug fix commit message is prefixed "fix:"

## Description

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/user/room.ts
+++ b/src/user/room.ts
@@ -428,12 +428,31 @@ export class Room extends Accessory implements Sayable {
     textOrContactOrFileOrUrl : string | Contact | FileBox | UrlLink,
     mention?                 : Contact | Contact[],
   ): Promise<void> {
+
+    const getRoomAlias: (c: Contact) => Promise<string> =
+    async (c: Contact) => {
+      try {
+        const alias: string | null = await c.alias()
+        return alias ? alias : c.name()
+      } catch (e) {
+        log.error('Room', 'say(%s,%s) get menction roomAlias error',
+                                        textOrContactOrFileOrUrl,
+                                        c.name())
+        return c.name()
+      }
+    }
+
+    const mentionAlias: string = Array.isArray(mention)
+                                  ? mention.map(async (c) => {
+                                      const roomAlias = await getRoomAlias(c)
+                                      return roomAlias
+                                    }).join(', ')
+                                  : mention ? await getRoomAlias(mention) : ''
+
     log.verbose('Room', 'say(%s, %s)',
                                   textOrContactOrFileOrUrl,
-                                  Array.isArray(mention)
-                                  ? mention.map(c => c.name()).join(', ')
-                                  : mention ? mention.name() : '',
-                )
+                                  mentionAlias)
+
     let text: string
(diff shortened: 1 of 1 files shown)


src/user/room.ts
```diff
@@ -428,12 +428,31 @@ export class Room extends Accessory implements Sayable {
     textOrContactOrFileOrUrl : string | Contact | FileBox | UrlLink,
     mention?                 : Contact | Contact[],
   ): Promise<void> {
+
+    const getRoomAlias: (c: Contact) => Promise<string> =
+    async (c: Contact) => {
+      try {
+        const alias: string | null = await c.alias()
```
zixia: Why you are using `c.alias()` here?

I believe what you really want is `room.alias(c)`

src/user/room.ts
```diff
@@ -428,12 +428,31 @@ export class Room extends Accessory implements Sayable {
...
+        log.error('Room', 'say(%s,%s) get menction roomAlias error',
+                                        textOrContactOrFileOrUrl,
+                                        c.name())
+        return c.name()
+      }
+    }
+
+    const mentionAlias: string = Array.isArray(mention)
+                                  ? mention.map(async (c) => {
+                                      const roomAlias = await getRoomAlias(c)
```
zixia: How about let's save `mention` to a `mentionContactList` like:

```
const mentionContactList = [...mention]
```

Then could we use 

```
const mentionNameList = await Promise.all(
  mentionContactList.map(
    c => this.alias(c) || c.name()
  )
)
``` 

instead of calling `getRoomAlias(c)`?

kis87988:
I will re-fork and re-do the commit since I using merge, and I tried to clean it but keep facing error. 
Let's do the code part first, after we done the code. I will create another PR for new fork.
Is this work?

src/user/room.ts
kis87988: yes... I forget the new version of room alias.

src/user/room.ts
kis87988: but if we using Promise.all() we might get (string|null) array for mentionNameList right?

kis87988: image

zixia: Yes, so you need another `.filter()` to filter those `null` values out.

zixia:
Please resolve all the reviews that I had posted, and it's ok to do the code part first.

src/user/room.ts
kis87988: Yes, I did try that...but ..
image

zixia: Yes, you can not rely on my pseudo code: you have to write it by yourself, good luck!

kis87988:
For clean commit, 
I will close this PR and continue on https://github.com/Chatie/wechaty/pull/1607

zixia:
Instead of close this PR, the better way is to fix your branch without close it, because it has many important information such as our discussion based on it.

PR #1607, opened 2018-09-28
fix: room mention contact should using roomAlias https://github.com/Chatie/wechaty/issues/1604

kis87988:
This is a PR continue with https://github.com/Chatie/wechaty/pull/1605 with clean commit
## I'm submitting a...

```
[ ] Bug Fix
[x] Feature
```

## Checklist

## Description

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/user/room.ts
+++ b/src/user/room.ts
@@ -429,22 +429,20 @@ export class Room extends Accessory implements Sayable {
     mention?                 : Contact | Contact[],
   ): Promise<void> {
 
-    const replyToList: Contact[] = [].concat(mention as any || [])
+    const replyToList: Contact[] = Array.isArray(mention)
+                                    ? [...mention]
+                                    : mention ? [mention] : []
 
-    const mentionAliasPromiseList = replyToList.length > 0
-                                    ? await Promise.all(
+    const mentionAliasPromiseList = await Promise.all(
                                       replyToList
                                       .map(c => this.alias(c) || c.name()))
-                                    : []
 
     const mentionAliasList: string[] = mentionAliasPromiseList
                                         .filter(a => a !== null) as string[]
 
     log.verbose('Room', 'say(%s, %s)',
                                   textOrContactOrFileOrUrl,
-                                  mentionAliasList.length > 1
-                                  ? mentionAliasList.join(', ')
-                                  : mentionAliasList.length === 1 ? mentionAliasList[0] : '')
+                                  mentionAliasList.join(', '))
 
     let text: string
 
--- a/src/user/room.ts
+++ b/src/user/room.ts
@@ -428,12 +428,31 @@ export class Room extends Accessory implements Sayable {
     textOrContactOrFileOrUrl : string | Contact | FileBox | UrlLink,
     mention?                 : Contact | Contact[],
   ): Promise<void> {
+
+    const getRoomAlias: (c: Contact) => Promise<string> =
+    async (c: Contact) => {
+      try {
+        const alias: string | null = await this.roomAlias(c)
+        return alias ? alias : c.name()
+      } catch (e) {
+        log.error('Room', 'say(%s,%s) get menction roomAlias error',
+                                        textOrContactOrFileOrUrl,
+                                        c.name())
+        return c.name()
+      }
+    }
+
+    const mentionAlias: string = Array.isArray(mention)
+                                  ? mention.map(async (c) => {
+                                      const roomAlias = await getRoomAlias(c)
+                                      return roomAlias
+                                    }).join(', ')
+                                  : mention ? await getRoomAlias(mention) : ''
+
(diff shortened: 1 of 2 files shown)


src/user/room.ts
```diff
@@ -428,23 +428,33 @@ export class Room extends Accessory implements Sayable {
     textOrContactOrFileOrUrl : string | Contact | FileBox | UrlLink,
     mention?                 : Contact | Contact[],
   ): Promise<void> {
+
+    const replyToList: Contact[] = [].concat(mention as any || [])
```
zixia: use `[...mention]` should be better.

src/user/room.ts
```diff
@@ -428,23 +428,33 @@ export class Room extends Accessory implements Sayable {
...
+
+    const mentionAliasPromiseList = replyToList.length > 0
```
zixia: I believe `replyToList.length > 0` is not necessary at here.

Rename `mentionAliasPromiseList` to `mentionAliasList` because it's not a promise.

src/user/room.ts
```diff
@@ -428,23 +428,33 @@ export class Room extends Accessory implements Sayable {
...
+                                        .filter(a => a !== null) as string[]
+
     log.verbose('Room', 'say(%s, %s)',
                                   textOrContactOrFileOrUrl,
-                                  Array.isArray(mention)
-                                  ? mention.map(c => c.name()).join(', ')
-                                  : mention ? mention.name() : '',
-                )
-    let text: string
+                                  mentionAliasList.length > 1
```
zixia: `mentionAliasList.length > 1` is not necessary at here.

src/user/room.ts
kis87988: but the replyToList is original code, and what if mention is undefine?

zixia: `[...(undefined||[])]`

src/user/room.ts
kis87988: mentionAliasPromiseList may contain ```string | null```, mentionAliasList should only contain ```string```. Therefore, I think the line 440 is the one being filter. 
However, you are right. I should give a good name for the one contain ```string | null```

src/user/room.ts
kis87988: mentionAliasPromiseList may contain ```string | null```, mentionAliasList should only contain ```string```. Therefore, I think the line 440 is the one being filter. 
However, you are right. I should give a good name for the one contain ```string | null```

src/user/room.ts
```diff
@@ -428,23 +428,31 @@ export class Room extends Accessory implements Sayable {
     textOrContactOrFileOrUrl : string | Contact | FileBox | UrlLink,
     mention?                 : Contact | Contact[],
   ): Promise<void> {
+
+    const replyToList: Contact[] = Array.isArray(mention)
```
zixia: Change this line to:

```
const replyToList: Contact[] = [...(mention || [])]
```

src/user/room.ts
```diff
@@ -428,23 +428,31 @@ export class Room extends Accessory implements Sayable {
...
+                                    ? [...mention]
+                                    : mention ? [mention] : []
+
+    const mentionAliasPromiseList = await Promise.all(
```
zixia: Rename `mentionAliasPromiseList` to `mentionAliasList`

src/user/room.ts
```diff
@@ -428,23 +428,31 @@ export class Room extends Accessory implements Sayable {
...
+                                      replyToList
+                                      .map(c => this.alias(c) || c.name()))
+
+    const mentionAliasList: string[] = mentionAliasPromiseList
```
zixia: This line can be removed because the array should not include the null anymore because of the above code `this.alias(c) || c.name()`

kis87988: but it is possible promise fail and got null value for ```c.name()```, isn't it?

kis87988: Before rename: 
image

zixia: `c.name()` will always return a `string`, so it should never be a `null`.

However, you can keep this code and I'll have a look after we merged it.

src/user/room.ts
kis87988: I don't think that works because the ```mention``` may be  ```Contact```, and I believe we don't want to have ```list of list(2D array)```  here
image

src/user/room.ts
kis87988: ok.

src/user/room.ts
kis87988: since below discussion, we should give a better name since that is await. 
However, ```mentionAliasList``` contain ```string | null``` base on ```Promise.all()``` fail will return ```null```. 
Therefore, I suggest name should be ```mentionAliasListResult``` or ```mentionAliasListPromiseResult```, but the second one name too long..

Issue #1642, opened 2018-11-15
Upgrade wechaty-puppet-padchat stable version


windmemory:
`wechaty-puppet-padchat` minor version 0.17 has been published for a while, and some fixes are pushed into this version. `v0.18.1` just got published, which we think should be a stable version for people to use. So please update the `wechaty-puppet-padchat` version in `puppet-config` to `v0.18.1`

zixia:
Merged.

windmemory:
Closed :)

PR #1643, opened 2018-11-15
Upgrade wechaty-puppet-padchat version in puppet config

windmemory:
Refer to https://github.com/Chatie/wechaty/issues/1642


windmemory:
Thanks! U r the best

Issue #1728, opened 2019-03-28
Recalled message design


windmemory:
**Is your feature request related to a problem? Please describe.**
Add new message function for recalled message

**Describe the solution you'd like**
The design would be like below
### For recalled message
* It will have a different message id from the original message
* The type for the message will be `Recall`
* There will be a new method named `recalled` to be added to the `Message` class and used to retrieve the original message

**Describe alternatives you've considered**
No alternatives :P

huan:
Related to: https://github.com/Chatie/wechaty-puppet-puppeteer/issues/31#issuecomment-470789051

@windmemory Please feel free to connect other related issues as well.

huan:
@windmemory Please close this issue if there's no more work needs to do on it.

windmemory:
Close this issue since the feature is implemented. Thanks for the help @huan

PR #1735, opened 2019-04-08
Add `Message.recalled()`

windmemory:
Refer to https://github.com/Chatie/wechaty/issues/1728


src/user/common.spec.ts
```diff
@@ -0,0 +1,13 @@
+import { SinonSandbox } from 'sinon'
+import { Puppet, RoomPayload } from 'wechaty-puppet'
+
+export const mockRoomPayload = (
```
huan: I do not think coupling between unit tests is a good idea, and also it seems no strong reason for us to seperate this method to create new file.

Please move the code to the unit test file instead of creating this new file.

src/user/message.ts
```diff
@@ -387,6 +387,24 @@ export class Message extends Accessory implements Sayable {
     return this.payload.text || ''
   }
 
+  public async recalled (): Promise<Message | null> {
```
huan: Please rename `recalled()` to `toRecalled()` because we have already had the following methods:

1. `toContact()`
1. `toFileBox()`
1. `toUrlLink()`

src/user/common.spec.ts
windmemory: The reason I do this is to make code climate happy, it was complaining about duplicate code in code base. I can revert the last commit if you think that's what you want.

src/user/message.ts
windmemory: Okay, will do.

windmemory:
@huan Please let me know if you want me to revert back the change of coupling the unit test or not. I will submit changes after that is confirmed :)

huan:
@windmemory Yes, please follow my reviews from your PR.

About the code climate I want to ignore it for the unit tests for a while, and think about it later, maybe config it to ignore all *.spec.ts code files.

windmemory:
Got it, thanks for the confirmation, will submit changes later.

windmemory:
Changed code according to review, and also add example code for `Message.toRecalled()`

Please let me know if you think anything else needs to be changed.

src/user/message.ts
```diff
@@ -387,6 +387,38 @@ export class Message extends Accessory implements Sayable {
...
+   */
+  public async toRecalled (): Promise<Message | null> {
+    if (this.type() !== MessageType.Recalled) {
+      throw new Error('Can not call toRecalled() on message which is not recalled type.')
+    }
+    const originalMessageId = this.text()
+    if (!originalMessageId) {
+      throw new Error('Can not find recalled message')
+    }
+    const message = this.wechaty.Message.load(originalMessageId)
```
huan: move the above line to inside `try {}` block.

huan:
Thanks for your great contribution!

windmemory:
np

Issue #1756, opened 2019-04-22
Missing example code


linyimin-bupt:
Versions
What is your wechaty version?
Answer:
mater branch

Describe the bug
Missing comment example code 

The `say` method supports `UrlLink`, but there is no example code.

PR #1757, opened 2019-04-22
add example code

linyimin-bupt:
refer #1756


src/user/message.ts
```diff
@@ -469,6 +469,18 @@ export class Message extends Accessory implements Sayable {
    *   }
    *
    * })
+   *
+   * // 4. send Link
```
windmemory: This piece of code should be in the bracket below.

huan:
@windmemory Thanks for the review!
@linyimin-bupt Good job!

Issue #1823, opened 2019-07-22
Would like to have receiver, serialize and deserialize functions on RoomInvitation


windmemory:
**Is your feature request related to a problem? Please describe.**
Trying to implementing a feature that user could accept room invitation and view room invitation in web page. So I need two more functions:
* One function enables the ability to retrieve/regenerate the room invitation, and accept when the user triggered the accept action.
* The other one could return back the invitation's receiver, for invitation that is not sent by bot, it is clear that the sender and the receiver, but if the invitation is sent by myself, it is impossible currently to know whom the invitation is sent to. So I want to have this function that tells me who is the receiver.

**Describe the solution you'd like**
I would like to have two or three more functions provide by `RoomInvitation` class. 
* `to(): Promise<Contact>` to retrieve the receiver info
* `stringify(): Promise<string>` to serialize the `RoomInvitation`
* `static parse(): Promise<RoomInvitation>` to deserialize the string and generate the room invitation object again, so we can call `accept` on it to accept the invitation.

**Describe alternatives you've considered**
No alternatives yet considered.

windmemory:
Kindly ping @huan

huan:
Sorry for the late reply, and thanks for the reminding.

The basic concept looks good in your proposal.

Please go-ahead to submit a PR to show your code so that we could be able to discuss the design patterns for it.

windmemory:
I've created a draft PR in `wechaty`, `wechaty-puppet` and `wechaty-puppet-padpro`. I added two class methods and two instance methods:
**Class Method**
* `serialize (invitation: RoomInvitation): Promise<string>`
* `deserialize<T extends typeof RoomInvitation> (payloadString: string): Promise<T['prototype']>`

**Instance Method**
* `to (): Promise<Contact>`
* `avatar (): Promise<FileBox>`

The class methods used to serialize the object back and forth so we can persist it in other storage, and load it, accept it anytime we want.

The instance methods used to get two more attributes for room invitation, which is the `to`, whom the invitation is sent to, and the `avatar` for the room that the invitation is for.

With the current functions, we are able to construct the invitation card with the functions on `RoomInvitation`.

But this is just a draft api, please let me know your idea about this, and any suggestion is welcome.

huan:
Great to hear from you and see the API drafts!

I will look into it and get back to you later.

windmemory:
Kindly ping @huan

huan:
[...]

PR #1828, opened 2019-07-29
Add more functions for room invitation class

windmemory:
Refer to https://github.com/Chatie/wechaty/issues/1823

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,7 +191,7 @@ All wechat contacts(friends/non-friends) will be encapsulated as a Contact.
 | static | [`load(query: string): Contact`](https://chatie.io/wechaty/#Contact.load) | get contact by id |
 | property | `id: readonly string` | get contact id |
 | method | [`sync(): Promise<void>`](https://chatie.io/wechaty/#Contact+sync) | force reload data for contact , sync data from lowlevel API again|
-| method | [`say(text: string): Promise<void>`](https://chatie.io/wechaty/#Contact+say) | send text, Contact, or file to contact |
+| method | [`say(text: string): Promise<void | Message>`](https://chatie.io/wechaty/#Contact+say) | send text, Contact, or file to contact, return the message which the bot sent (only `puppet-padplus` supported). |
 | method | [`self(): boolean`](https://chatie.io/wechaty/#Contact+self) | check if contact is self |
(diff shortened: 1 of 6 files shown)


src/user/room-invitation.ts
```diff
@@ -49,6 +51,18 @@ export class RoomInvitation extends Accessory implements Acceptable {
     return newRoomInvitation
   }
 
+  public static async serialize (invitation: RoomInvitation): Promise<string> {
```
huan: See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#Description
> If the value has a toJSON() method, it's responsible to define what data will be serialized.

Please follow the standard JavaScript design by renaming `serialize` to `toJSON()`

src/user/room-invitation.ts
```diff
@@ -49,6 +51,18 @@ export class RoomInvitation extends Accessory implements Acceptable {
...
+    return this.puppet.roomInvitationSerialize(invitation.id)
+  }
+
+  public static async deserialize<T extends typeof RoomInvitation> (
```
huan: `deserialize` should be renamed to `fromJSON`

src/user/room-invitation.ts
```diff
@@ -241,4 +255,19 @@ export class RoomInvitation extends Accessory implements Acceptable {
...
+
+    const payload = await this.puppet.roomInvitationPayload(this.id)
+    const to = this.wechaty.Contact.load(payload.toId)
+    return to
+  }
+
+  public async avatar (): Promise<FileBox> {
+    const payload = await this.puppet.roomInvitationPayload(this.id)
+    const fileBox = FileBox.fromUrl(payload.avatar, `${payload.roomTopic}_avatar.jpg`)
+    await fileBox.ready()
```
huan: What's the reason we need a `fileBox.ready()` at here?

src/user/room-invitation.ts
windmemory: sure, will do.

src/user/room-invitation.ts
windmemory: Because this is a `FileBox` from url, the content is not loaded if no `ready` is called.

Perviously I've met some error thrown when calling some function on `FileBox`, so I get used to adding `ready` function before use the `FileBox`.

Issue #1829, opened 2019-07-29
Would like to have the timestamp on the room/friendship events


windmemory:
**Is your feature request related to a problem? Please describe.**
Currently we don't have a way to get the timestamp of when the events really happened. In wechat's world, when the account is not online, the event is queued up, and once the account is online, events will be pushed. If we use the timestamp that the event is triggered, it is not accurate, sometimes, it will have few hours or even days delay. So I would like to have the timestamp value emitted together in the room/friendship events.

**Describe the solution you'd like**
In the listener function of room/friendship functions, add `timestamp` as the last parameter.

**Describe alternatives you've considered**
No alternatives considered.

**Additional context**
No

[enhancement]

huan:
It's a reasonable proposal, please go-ahead and PR will be welcome!

windmemory:
Okay will do

windmemory:
I've created draft changes for this feature request.

Actually I came up with two solutions:
* Add timestamp value to the event callbacks
* Convert the arguments in the event callback into one event object, then add timestamp onto that object

I took the first one, since the second one requires more code changes than the first one, and that is a big breaking change.
But for the first solution, my PR might not be the best solution. At the beginning, I thought to put the new arguments at the end of the parameters, so this change won't be a breaking change, but I found the `room-leave` event has an optional parameter, and from my understanding, the new `time` argument should always be available, so it is not optional, so I can not put it at the end of every event callback. So I put it after the `room` argument.

Let me know what do you think is the best implementation. @huan

windmemory:
I think we can close this issue as soon as the `wechaty-puppet-puppeteer` update the code to emit the `timstamp` in room events.

huan:
You are right. And not only the wechaty-puppet-puppeteer but also other puppet providers maintained by chatie as well, like `mock`, `wechat4u`, etc.

windmemory:
Issues created in each puppet, I will try to squeeze some time working on some of the upgrades.

huan:
Great, thank you very much for creating the issues and planing to fix them!

huan:
After we had close this issue, I'd like to bump all the puppets version to a stable version number, so that all the developers would be able to use them.

Especially the puppeteer and the padpro.

huan:
I will close this issue because I guess it has been resolved.

PR #1832, opened 2019-08-05
Add timestamp to events

windmemory:
Refer to https://github.com/Chatie/wechaty/issues/1829


src/user/room.ts
```diff
@@ -543,10 +543,10 @@ export class Room extends Accessory implements Sayable {
     }
   }
 
-  public emit (event: 'invite', inviter: Contact,         invitation: RoomInvitation)           : boolean
-  public emit (event: 'leave',  leaverList:   Contact[],  remover?: Contact)                    : boolean
-  public emit (event: 'join',   inviteeList:  Contact[],  inviter:  Contact)                    : boolean
-  public emit (event: 'topic',  topic:        string,     oldTopic: string,   changer: Contact) : boolean
+  public emit (event: 'invite', inviter: Contact,         invitation: RoomInvitation)                       : boolean
+  public emit (event: 'leave',  time: Date, leaverList:   Contact[],  remover?: Contact)                    : boolean
```
huan: Please move all the `time` argument to the end of the argument lists, and make it optional, so that it will not become a breaking change.

huan: And also, please change the `remover?` argument for the `emit` function to `remover`. I think it's better to not optional as a emit argument.

src/wechaty.ts
```diff
@@ -311,19 +311,19 @@ export class Wechaty extends Accessory implements Sayable {
...
-  public on (event: 'room-invite',  listener: string | ((this: Wechaty, roomInvitation: RoomInvitation) => void))                                   : this
-  public on (event: 'room-join',    listener: string | ((this: Wechaty, room: Room, inviteeList: Contact[],  inviter: Contact) => void))            : this
-  public on (event: 'room-leave',   listener: string | ((this: Wechaty, room: Room, leaverList: Contact[], remover?: Contact) => void))             : this
-  public on (event: 'room-topic',   listener: string | ((this: Wechaty, room: Room, newTopic: string, oldTopic: string, changer: Contact) => void)) : this
-  public on (event: 'scan',         listener: string | ((this: Wechaty, qrcode: string, status: ScanStatus, data?: string) => void))                : this
-  public on (event: 'start' | 'stop', listener: string | ((this: Wechaty) => void))                                                                 : this
+  public on (event: 'dong',         listener: string | ((this: Wechaty, data?: string) => void))                                                                : this
```
huan: Please do not include unrelated changes.

src/wechaty.ts
```diff
@@ -683,13 +683,13 @@ export class Wechaty extends Accessory implements Sayable {
             const inviter = this.Contact.load(inviterId)
             await inviter.ready()
 
-            this.emit('room-join', room, inviteeList, inviter)
-            room.emit('join', inviteeList, inviter)
+            this.emit('room-join', room, new Date(timestamp), inviteeList, inviter)
```
huan: save `new Date(timestamp)` to a variable and reuse it instead of call `new Date` twice.

src/user/room.ts
windmemory: Will do.

src/wechaty.ts
windmemory: This is aligning the `dong` event to all the other events, since this change modified the indentation of other events, is this necessary?

huan: Not necessary for this PR, because we have to focus on the code design.

We can fix the align after this PR had been merged by another PR, or we add the align as the lastest commit after this PR had been approved and ready to be merged.

src/wechaty.ts
windmemory: I would prefer to have different variables, since users might modify the value of the variable(you never know).
If they listened on `wechat`'s `room-join` events, also listened to the specific `room`'s `room-join` event, modifying one time will have effect on the other one, which might cause some very unexpected bug in the future.

Alternative solution: we could make this value readonly or freeze this time variable.

What do you think about this?

src/wechaty.ts
windmemory: okay, will revert this change.

huan:
@windmemory Could you please mark my reviews as `resolved` after you resolve it? Or they will keep showing and it will not very clear for our progress. Thanks.

src/user/room.ts
```diff
@@ -544,9 +544,9 @@ export class Room extends Accessory implements Sayable {
   }
 
   public emit (event: 'invite', inviter: Contact,         invitation: RoomInvitation)           : boolean
-  public emit (event: 'leave',  leaverList:   Contact[],  remover?: Contact)                    : boolean
-  public emit (event: 'join',   inviteeList:  Contact[],  inviter:  Contact)                    : boolean
-  public emit (event: 'topic',  topic:        string,     oldTopic: string,   changer: Contact) : boolean
+  public emit (event: 'leave',  leaverList:   Contact[],  remover:  Contact, time?: Date)                    : boolean
```
huan: The `time` should be optional for `emit` because we must make sure if there's some consumer want to get the time, they will.

Remove `?` to make sure the `time` will be passed.

And I also feel that the `time` should rename to `date` because it's a `Date`.

src/wechaty.ts
```diff
@@ -295,9 +295,9 @@ export class Wechaty extends Accessory implements Sayable {
   public emit (event: 'message',    message: Message)                                                   : boolean
   public emit (event: 'ready')                                                                          : boolean
   public emit (event: 'room-invite',  roomInvitation: RoomInvitation)                                   : boolean
-  public emit (event: 'room-join',    room: Room, inviteeList : Contact[], inviter  : Contact)          : boolean
-  public emit (event: 'room-leave',   room: Room, leaverList  : Contact[], remover? : Contact)          : boolean
-  public emit (event: 'room-topic',   room: Room, newTopic: string, oldTopic: string, changer: Contact) : boolean
+  public emit (event: 'room-join',    room: Room, inviteeList : Contact[], inviter : Contact, time?: Date)           : boolean
+  public emit (event: 'room-leave',   room: Room, leaverList  : Contact[], remover : Contact, time?: Date)           : boolean
```
huan: 1. Should be optional
1. Consider rename it to `date`

src/wechaty.ts
```diff
@@ -683,27 +683,24 @@ export class Wechaty extends Accessory implements Sayable {
             const inviter = this.Contact.load(inviterId)
             await inviter.ready()
 
-            this.emit('room-join', room, inviteeList, inviter)
-            room.emit('join', inviteeList, inviter)
+            this.emit('room-join', room, inviteeList, inviter, new Date(timestamp))
```
huan: Create a `date` var to store the `new Date()` value. 

Then you can use it for twice.

src/wechaty.ts
```diff
@@ -683,27 +683,24 @@ export class Wechaty extends Accessory implements Sayable {
...
-            if (removerId) {
-              remover = this.Contact.load(removerId)
-              await remover.ready()
-            }
+            const remover = this.Contact.load(removerId)
+            await remover.ready()
 
-            this.emit('room-leave', room, leaverList, remover)
-            room.emit('leave', leaverList, remover)
+            this.emit('room-leave', room, leaverList, remover, new Date(timestamp))
```
huan: Create a temp variable from here as well.

src/wechaty.ts
huan: I don't think this will happen because the user can only modify their local copy of the variable value.

src/user/room.ts
windmemory: actually it is a `datetime`, shall we rename it as `datetime`?

windmemory:
Revised code according to review and one comment to discuss.

src/user/room.ts
```diff
@@ -583,9 +583,9 @@ export class Room extends Accessory implements Sayable {
   }
 
   public on (event: 'invite', listener: (this: Room, inviter: Contact, invitation: RoomInvitation) => void)               : this
-  public on (event: 'leave',  listener: (this: Room, leaverList:  Contact[], remover?: Contact) => void)                  : this
-  public on (event: 'join',   listener: (this: Room, inviteeList: Contact[], inviter:  Contact) => void)                  : this
-  public on (event: 'topic',  listener: (this: Room, topic:       string,    oldTopic: string, changer: Contact) => void) : this
+  public on (event: 'leave',  listener: (this: Room, leaverList:  Contact[], remover:  Contact, date: Date) => void)                  : this
```
huan: We have to keep the `remover?` as optional or we will introduce breaking changes to our library.

And so that you will have to make the `date?` optional too.

src/wechaty.ts
```diff
@@ -322,9 +322,9 @@ export class Wechaty extends Accessory implements Sayable {
...
   public on (event: 'ready',        listener: string | ((this: Wechaty) => void))                                                                   : this
   public on (event: 'room-invite',  listener: string | ((this: Wechaty, roomInvitation: RoomInvitation) => void))                                   : this
-  public on (event: 'room-join',    listener: string | ((this: Wechaty, room: Room, inviteeList: Contact[],  inviter: Contact) => void))            : this
-  public on (event: 'room-leave',   listener: string | ((this: Wechaty, room: Room, leaverList: Contact[], remover?: Contact) => void))             : this
-  public on (event: 'room-topic',   listener: string | ((this: Wechaty, room: Room, newTopic: string, oldTopic: string, changer: Contact) => void)) : this
+  public on (event: 'room-join',    listener: string | ((this: Wechaty, room: Room, inviteeList: Contact[], inviter: Contact, date?: Date) => void))             : this
+  public on (event: 'room-leave',   listener: string | ((this: Wechaty, room: Room, leaverList: Contact[],  remover: Contact, date?: Date) => void))             : this
```
huan: optional.

I would like to make the `remover` not optional, in the next major version upgrade.

windmemory:
That's right, changed according to review suggestions.

package.json
```diff
@@ -98,7 +98,7 @@
     "semver": "^6.0.0",
     "state-switch": "^0.6.2",
     "watchdog": "^0.8.1",
-    "wechaty-puppet": "^0.15.25",
+    "wechaty-puppet": "^0.15.26",
```
huan: Do you want to upgrade this version as well?

windmemory: Yes, I should have.

Bumped the version to `0.15.27`

huan:
Merged.

PR #1833, opened 2019-08-06
Update wechaty.ts

su-chang:
fix: remove `room.sync()`, because it can not get room info when the bot has been removed from room. #1834

## I'm submitting a...

- [x] Bug Fix
- [ ] Feature
- [ ] Other (Refactoring, Added tests, Documentation, ...)

## Checklist

## Description

```
[ ] Yes
[ ] No
```


huan:
I don't think this fix is a good idea because you just fix your problem, but at the same time, you made the system worse.

The `room.sync()` was called on its purpose, your modification just like "if your code can not pass the unit test, then disable the unit test."

Please think more about how the whole system works, and find a better way to handle your problem. And also you have to think about what's the best practice to do it.

su-chang:
Thanks for pointing out this problem, I will try to fix this bug gracefully.

su-chang:
About the event `room-leave`:

**Case 1:** the bot is the room owner
-  the leaver (not friend)
if the bot start after the remove action, it can not get the leaver info, the params `leaverIdList` is an empty array (**not make sense**, user want to know which member the bot removed out)

- the leaver (friend)
the params `leaverIdList` is an array and its length is 1, in fact, when you remove some members in one room, WeChat will do it one by one, so the `leaverIdList`'s length always be 1 (**not make sense**, make Contact [] to Contact?)

**Case 2:** the bot is not the room owner
- the owner remove others
system will not notify the bot, so the bot has no response with this option (**make sense**)

- the owner remove the bot
the bot can not get the room info after been removed from one room (**not make sense**, user want to know which room the bot lost)

There are some cases that I summary for the event `room-leave`, I hope u can make a rule for it, just like which way is right. thx!

huan:
@su-chang Thanks for the summary.

So what's your suggestion to deal with this problem?

huan:
I guess this problem is because when we call `sync()` on the room that removed the bot, the puppet deleted all the room payload, which caused the room id can not get any payload data anymore.

I believe it will be easy to fix by keeping the room payload after the bot had been removed out.

We can think about the Wechat App on the phone: If someone is removed from a room(group), the Wechat App will keep the room information in the conversation list, instead of remove/delete it, even they can not get any information from that room anymore.

windmemory:
@huan Do you think we need a flag to tell the user that the room is deleted?

huan:
@su-chang Thanks for sharing your thoughts and the design.

I think Wechaty needs to do a `sync()` after the bot leaves a room, so I kept the `sync()` in wechaty.

However, I do agree with @windmemory that we should set a flag for the room to identify that the bot was not in it anymore, but we should keep the room information for future use. (that's what Wechat App on the Phone did).

So I'd like to suggest that we deal this in the puppet implementation, and we can set the out-of-the-room flag in the future, then the Wechaty will get to know it after the sync operation.

See: https://github.com/wechaty/wechaty-puppet-padplus/pull/119

PR #1859, opened 2019-09-29
fix opencollective Sponsors style

lijiarui:


huan:
Could you please describe that "what's the problem now, and what you are trying to fix", instead of just putting a title with "fix something"?

Thanks.

lijiarui:
Sure. What I did is trying to sync open-collective sponsor list like baker shows. I guess why the current code is PR-ed by the open collective themselves is because we don't have sponsor before.

### Before

image


### After

image

huan:
It looks great, thank you for the fixing and the sponsor on open collective!

BTW I believe you need only one name (juzi.bot) instead of two (botorange), which will be good for to your branding.

Issue #1856, opened 2019-09-24
Implement Label in Wechaty


su-chang:
**Is your feature request related to a problem? Please describe.**

if someone has a lot of friends, `Label` is indeed needed for manage friends list.

**Describe the solution you'd like**
we need some function like these: 
  - get all labels (For example: [{label: 'xxx', members: ['xxx1', 'xxx2', ... , 'xxxx']}])
  - add label
  - update friend's label

**Describe alternatives you've considered**

The function of `Label` is depended on puppet which one has supported.

- supported
  - macpro
  - padplus
- not supported
  - puppeteer
  - padpro

if the puppet is not supported, we can do nothing.

**Additional context**
I will create a PR for this issue later.

[enhancement]

huan:
Thank you very much for the proposal!

A PR will be welcome for discussing, and also please notice that we have another thread talked about this topic before.

Please link to that issue and also read it carefully to learn more from the past discussions.

su-chang:
#1627

su-chang:
#1627

huan:
Glad to see you had connected to the previous conversion.

It is very important that we should always continue our discussion, based on the already learned knowledge, instead of creating a branding new one from scratch.

su-chang:
@huan 
Here are some Raw API and their response data.

- Raw API name : AddLabel

> bot.Tag.newTag(tag: string)

```js
{
  labelList:
    [ { LabelName: '测试标签1', LabelID: 1 }, ],
  loginer: 'xxx',
  count: 1,
  message: '',
  status: 0,
}
```

- Raw API name : GetAllLabel
> bot.Tag.tagList()
```js
{
  labelList:
   [ { LabelName: '测试标签1', LabelID: 1 },
     { LabelName: '测试标签3', LabelID: 3 },
     { LabelName: '测试标签2', LabelID: 2 } ],
  loginer: 'xxx',
  count: 3,
  message: '',
  status: 0
}
```

- Raw API name : ModifyLabel
> contact.addTag(contactId:string, tag: string)
> we can get `contactId` by contact
```js
{
  loginer: 'xxx',
  labelIds: '1,3',
  message: '',
  userName: 'Soul001001',
  status: 0
}
```

su-chang:
For wechaty:

I think we should create a `Tag` class, the methods list:

- newTag(tag: string): Promise<string>
- findAll(): Promise<TagPayload []>
- find(tag: string): Promise<TagPayload>
[...]

PR #1864, opened 2019-10-11
Implement Label related methods

su-chang:
## I'm submitting a...

```
[ ] Bug Fix
[x] Feature
```

## Checklist

## Description

```
[ ] Yes
[ ] No
```


su-chang:
related issue: https://github.com/Chatie/wechaty/issues/1856

huan:
Please reference the related issue in your PR, because we need to track everything in the main issue.

Create it if it does not exist yet, and descript what you want to do in details in the issue body.

Thank you very much.

src/user/tag.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    const tag = await this.puppet.searchTag(query)
+
+    return tag
+  }
+
+  /**
+   * show all labels across the login account
+   */
+  public async tagList (): Promise<TagPayload[]> {
```
huan: What's the difference between the `tagList` and the `findAll`?

As my understanding, they will all return all the tags, so I'd like to suggest to remove the `tagList` method because it does not contribute any new value for us.

src/user/tag.ts
```diff
@@ -0,0 +1,86 @@
...
+    } catch (e) {
+      log.error('Tag', 'tagList() exception: %s', e.message)
+      throw new Error(`can not get tag list`)
+    }
+  }
+
+  /**
+   * create a new label
+   */
+  public async newTag (tag: string): Promise<string> {
```
huan: newTag -> create

I believe it will be better to make a call like `bot.Tag.create('myTag')` instead of `bot.Tag.newTag('myTag')`.

src/user/tag.ts
su-chang: Agree!

> findAll()  => find all tags from cache
> tagList() => find all tags from API search

You are right, we do not need to know the data source in Wechaty.

src/user/tag.ts
su-chang: Yes, you are right, `create` is better than `newTag`.

su-chang:
@huan I have implemented all Label related methods conform to design, waiting for your review and suggestions. Thank you.

src/user/tag.ts
```diff
@@ -0,0 +1,202 @@
...
+   * @example
+   * const bot = new Wechaty()
+   * await bot.start()
+   * await bot.Tag.get(tag)
+   */
+  public static async get<T extends typeof Tag> (this: T, tag: string): Promise<T['prototype']> {
+    log.verbose('Tag', 'get()')
+
+    try {
+      const tagPayload: TagPayload = await this.puppet.createTag(tag)
```
windmemory: If the tag exists, we should not create a new tag here, right?

su-chang: It will return the existed tag instead of creating.

windmemory: Then the function is misleading. We shouldn't call it `createTag`, it should be `getOrCreateTag` or something else. `createTag` explicitly tell the user that this is creating a tag.

su-chang: Agree, `getOrCreateTag()` is better for the puppet method.

windmemory: I guess @huan might not like this design :P

Let's wait for his review.

su-chang:
@huan  ping

huan:
I have re-write some code in this PR for designing a more clear & easy to use published API.

Please let me know what you think @su-chang @windmemory @xinbenlv , thank you very much!

src/user/tag.ts
```diff
@@ -0,0 +1,182 @@
...
+   * Delete this tag
+   *
+   * > Tips:
+   * This function is depending on the Puppet Implementation, see [puppet-compatible-table](https://github.com/Chatie/wechaty/wiki/Puppet#3-puppet-compatible-table)
+   *
+   * @returns {Promise<void>}
+   * @example
+   * await tag.del()
+   */
+  public async del (from: Contact): Promise<void> {
```
windmemory: If we want to delete a tag entirely, how to achieve that  with the current design?

src/user/tag.ts
```diff
@@ -0,0 +1,182 @@
...
+   * @example
+   * const bot = new Wechaty()
+   * await bot.Tag.get(tag)
+   */
+  public static async get<T extends typeof Tag> (
+    this: T,
+    tag: string,
+  ): Promise<T['prototype']> {
+    log.verbose('Tag', 'get(%s)', tag)
+    return this.load(tag)
```
windmemory: In the `get` function, there is no actual puppet call, then when shall we create the tag?

su-chang:
I have see these methods only maintain the relationship between Tag and Contact, but no method work for Tag. 

For example, if we want to remove one tag from tag list, we need `delete` method, so that we could never see this tag again in our WeChat App. If we do not have this method, the tags in our WeChat App will be mess really, and so does `modify`.

src/user/tag.ts
```diff
@@ -87,23 +87,30 @@ export class Tag  extends Accessory {
...
+    target: typeof Contact | typeof Favorite,
+  ): Promise<Array<T['prototype']>> {
+    log.verbose('Tag', 'findAll(%s)', target)
+
+    const tagList: Tag[] = []
+
+    if (target === Contact || target === this.wechaty.Contact) {
+      const contactList = await this.wechaty.Contact.findAll()
+      for (const contact of contactList) {
+        const contactTagList = await contact.tags()
```
su-chang: Do we need add method `tags()` in wechaty-puppet?

Or change it to this way:

```js
const contactTagList = await this.puppet.tagListContact(contact.id)
```

src/user/tag.ts
huan: We can not do delete tag itself operations in this design.

I think we can just leave tag as they are will be cleaner for the Tag API.

BTW: I think we should rename all `del()` to `remove()` because it's more intuitive?

src/user/tag.ts
huan: In my current design, I want to use Lazy Operation for the Tag.

in the `get()` we only save the tag string in the class. The actual operation will happen when we call `add()` or `del()`, and we pass the tag string to the puppet API, then puppet API deal with the tag name with the Contact or Favorite.

I think this will be a clear design with the high-level API, please feel free to let me know what you think.

src/user/tag.ts
su-chang: Cleaner Tag API will make Tags in WeChat not clean.

If we don't remove tag which we do not need any more, the tag will make WeChat Tags worse. I think this design will make our user who use tag frequently feel not well.

`remove()` is better for this case.

src/user/tag.ts
su-chang: Lazy Operation is perfect.

So we could use this method like this way : 
```js
const tag = await bot.Tag.get('myTag')
const friend =bot.Contact.load(xxx)
await tag.add(friend) // we will create a new tag named myTag in this method
```

So we need to check the tag exist or not each time when we use `add()`?

I have a question about it, how could we know the tag exist or not ? Or just create new tag each time when we use `add()`?

This design will make `tagAddContact()` a little heavy, it means `create()` + `add()`, if something wrong happened in `create()`, it will wrong many times.

BTW, this method should check the tag's length, if user input a string which length over 15, we can not give a right tag back to user. It will be exposed when they use `add()` method.

huan:
1. I'll consider adding a `delete()` method to clean the Tag for the system
1. I'll assume that the Tag length will not longer than 15 characters. (in UTF8)
1. I believe it will be better to let the puppet to decide if they need to create a tag before using it.

Any suggestions?

su-chang:
I'd love to the Lazy Operation, and `delete()` is great for Tag system.

Do we need a method which named `modify()`? 

It has been supported by the WeChat Protocol, and it will make the Tag system be better. But not must.

I'm not insist on it. Just a suggestion.

huan:
The Tag class will be available from Wechaty v0.29.41 or later.

PR #1886, opened 2019-12-09
Recall msg

SilentQianyi:
## I'm submitting a...

```
[ ] Bug Fix
[ ] Feature
```

## Checklist

## Description

```
[ ] Yes
[ ] No
```

Issue
https://github.com/wechaty/wechaty/issues/1885


src/user/message.ts
```diff
@@ -566,6 +567,32 @@ export class Message extends Accessory implements Sayable {
...
+   * Recall a message.
+   * > Tips:
+   * @returns {Promise<boolean>}
+   *
+   * @example
+   * const bot = new Wechaty()
+   * bot
+   * .on('message', async m => {
+   *   const recallMessage = await msg.say('123')
+   *   const isSuccess = await recallMessage.recall()
```
CooperFu: recall should need a messageId.

windmemory: I don't think so, this is a function on `Message` class, it is recalling the message which called this function.

windmemory: I don't think so, this is a function on `Message` class, it is recalling the message which called this function.

CooperFu: ok then, I send two messages. what should I do for I want to recall the first one?
```
const recallMessage = await msg.say('123')
const recallMessage2 = await msg.say('456')
const isSuccess = await recallMessage.recall()
```
like this?

SilentQianyi: yes.

src/user/message.ts
```diff
@@ -566,6 +566,24 @@ export class Message extends Accessory implements Sayable {
...
```
huan: `recallMessage` might be null, please fix the example code for this.

PR #1891, opened 2019-12-25
Delay friendship

SilentQianyi:
## I'm submitting a...

```
[ ] Bug Fix
[ ] Feature
```

## Checklist

## Description

```
[ ] Yes
[ ] No
```

Issus
https://github.com/wechaty/wechaty/issues/1890


src/user/friendship.ts
```diff
@@ -357,4 +357,68 @@ export class Friendship extends Accessory implements Acceptable {
...
+   * bot.on('friendship', async friendship => {
+   *   try {
+   *     const payload = await friendship.toJson()
+   *   } catch (e) {
+   *     console.error(e)
+   *   }
+   * }
+   * .start()
+   */
+  public async toJson (): Promise<string> {
```
huan: According to <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify>

> ### toJSON() behavior
> If an object being stringified has a property named toJSON whose value is a function, then the toJSON() method customizes JSON stringification behavior: instead of the object being serialized, the value returned by the toJSON() method when called will be serialized. JSON.stringify() calls toJSON with one parameter:
> 
> - if this object is a property value, the property name
> - if it is in an array, the index in the array, as a string
> - an empty string if JSON.stringify() was directly called on this object

The method name should be named properly as `toJSON()` with a return type `string` instead of a `Promise`.

PR #1894, opened 2019-12-26
Room say bug

SilentQianyi:
## I'm submitting a...

```
[ ] Bug Fix
[ ] Feature
```

## Checklist

## Description

```
[ ] Yes
[ ] No
```

Issue
https://github.com/wechaty/wechaty/issues/1893


huan:
As I mentioned in the issue https://github.com/wechaty/wechaty/issues/1893#issuecomment-569659199, I think there will never happen that the `textList.length !== (mentionList.length + 1)` when we are using the ```Room.say`` ``` ES6 call currectly.

So this code block will be never executed unless we are using some _hackly_ way, which is not recommended.

huan:
I'll close this issue because I believe that this condition is an anti-pattern that should not be supported by the framework.

A workaround to fulfill your requirements on this condition is that you can call the `wechaty.puppet.messageSend()` directly, so that you can do whatever you want.

PR #1895, opened 2019-12-27
Friend search

su-chang:
## I'm submitting a...

```
[ ] Bug Fix
[x] Feature
```

## Checklist

## Description

```
[ ] Yes
[ ] No
```


su-chang:
Related issue: https://github.com/wechaty/wechaty/issues/1892

src/user/friendship.ts
```diff
@@ -70,6 +71,40 @@ export class Friendship extends Accessory implements Acceptable {
...
+   *
+   */
+  public static async search (
+    condition: FriendshipSearchCondition
+  ): Promise<Contact> {
+    log.verbose('Friendship', 'static search(%s)',
+      condition.phone || condition.weixin
+    )
+    if (condition.phone || condition.weixin) {
+      const id = condition.phone || condition.weixin || ''
```
windmemory: We don't need the `|| ''` since there is a check on the `condition.phone` and `condition.weixin`

src/user/friendship.ts
```diff
@@ -70,6 +71,40 @@ export class Friendship extends Accessory implements Acceptable {
...
+      await this.puppet.friendshipSearch(id)
```
windmemory: Shall we return a value for this function call?

src/user/friendship.ts
su-chang: Of course, but without empty string the id will be regard as undefined type, and then we have to use `id!`, which way is better? Or any other better way?

huan:
This feature will be available from v0.29.40 or later.

Thank you @su-chang @windmemory for adding this feature to Wechaty!

PR #1900, opened 2020-01-06
Add reason for logout event

su-chang:
## I'm submitting a...

```
[ ] Bug Fix
[x] Feature
```

## Checklist

## Description

```
[ ] Yes
[ ] No
```


su-chang:
Related issue: https://github.com/wechaty/wechaty-puppet-padplus/issues/100

su-chang:
I think optional is flexible, if puppet which do not support the logout reason, it could ignore the parameter `reason`.

windmemory:
If no reason, an empty string could also work. But optional also works. I will not insist on this point.

But I would insist on the separation of `login` and `logout` event since there is no reason for login event right now.

su-chang:
Got it.

huan:
@windmemory We should not approve any PR that did not pass the CI. If the CI is not green, we should never approve it unless we have to. Thanks!

su-chang:
@huan ping

huan:
Thanks for making the CI green for this PR.

PR #1903, opened 2020-01-13
Room announce bug

SilentQianyi:
Issue
https://github.com/wechaty/wechaty/issues/1902


windmemory:
Please fix the CI build.

huan:
We should not add another `null` parameter to do this task.

We should use an empty string as the announcement to clean the announcement from a room, which is more clear.

I'll close this PR for now, and please let me know if you have any questions.

PR #1946, opened 2020-04-18
feat: added wechaty plugin

Gcaufy:
## Wechaty.use

```
[ ] Bug Fix
[x] Feature
```

## Checklist

## Description

https://github.com/wechaty/wechaty/issues/1939

We can discuss it here.

```
[ ] Yes
[x] No
```


Gcaufy:
As the discussion here: https://github.com/wechaty/wechaty/issues/1939

I made few changes of the demo code. Just want to make sure you are agree with it.
So I can continue on the rest

### 1. I would like to call it `Plugin` instead of `MiddleWare`

In my opinion, middleware is alway dealing with the input/output.  Middleware A deal with it, and then decide to handover to Middleware B. But we are not doing that. We just installed it, and then it do works for you. It's more like a `Plugin`.

I think a MiddleWare is something like this:
```
Wechaty.use({ message: function (message: string, next: (message: string) => void) {
  if (message.indexOf('hi') > -1) {
     next();
  }
})
```
So that's the reason I change the name to `Plugin`. and implement it like this.

### 2. I would prefer to use a abstract class as a Plugin instead of a plugin function.
```
// 1. Plugin function
const MyPlugin = function (options: any) {
  return function(this: Wechaty) {
    this.on('join-room', () => console.log('joined'));
  }
}
Wechaty.use(MyPlugin({a: 1}));

// 2. Plugin Class
Class MyPlugin {
  constructor(options) { this.options = options }
  install(bot: Wechaty) {
    bot.on('join-room', () => console.log('joined'));
  }
}
Wechaty.use(new MyPlugin({a: 1}));
```
Basically it does the same thing. but more readable, and easier to understand. 

### 3. I would prefer the `use` function in Class instead of Instance.

Which is:
```
Wechaty.use(new MyPlugin());
```

but I made it both support the Class and the instance. also support the Array
```
Wechaty.use(new CommonPlugin());

const bot = new Wechaty();
bot.use([ new MyOwnPlugin(), new MyAnotherPlugin() ]);
```

huan:
Hi @Gcaufy ,

Thank you for the design of our powerful Middleware/Plugin system for Wechaty!

I have read your questions and also the source code in the PR, I believe we are in the right way to archive our goal.

Here's my reply for your 3 questions:

> 1. Call it Plugin instead of MiddleWare

I have no preference between those two names because I feel they are all fit our design.

I agree with you that the middleware mostly they are chaining for doing something together, but in our system, it does not. But the name of "Middleware" sounds more fashionable.

And the name of "Plugin" is more classic, simple, and clean, so they are all good.

Please feel free to make your decision about selecting the name, and I'll support your final choice!

> 2. Plugin Interface: an abstract class, or a plugin function?

My opinion for the Interface is not to use a Class because it looks too heavy.

I'd like to purpose the following design, to archive the same goals:

```ts
export type WechatyPlugin = (...options: any[]) => (wechaty: Wechaty) => void

const kickPlugin: WechatyPlugin = (managedRoomTopic: string) => wechaty => {
  wechaty.on('message', message => {
    ...
  })
}

wechaty.use(kickPlugin('My Room Topic'))
```

As my personal feeling, the above `Functional` design will be shorter and clearer than the `Class` design from yours.

So I'd like to purpose that our Plugin interface would be better to be defined as a Functional Interface.

> 3. to define `use()` as a Class static method instead of the Instance method

Could you please explain it with more examples about why you want to design like that?

I feel it will be more clear and simple if we only provide a `use()` method for the Wechaty instance so that every instance is separated by default.

-----

I'll also add some comments to the PR code, and please feel free to let me know what you think.

Thank you for your design and questions again, and I hope you have a great weekend!

src/wechaty.ts
```diff
@@ -114,6 +114,15 @@ export interface WechatyOptions {
   ioToken?       : string,                    // Io TOKEN
 }
 
+// A plugin is a Class with a install function.
+export abstract class WechatyPlugin {
```
huan, line 118: How about a `Functional` interface to archive the same goal?

I'd like to purpose this interface:

```ts
export type WechatyPlugin = (...options: any[]) => (wechaty: Wechaty) => void
```

src/wechaty.ts
```diff
@@ -152,6 +161,10 @@ export class Wechaty extends EventEmitter implements Sayable {
    */
   private static globalInstance: Wechaty
 
+  private static globalPlugins: WechatyPlugin[] = []
```
huan, line 164: I'd like to suggest that we separate the effects between the Wechaty Instances, so that every Wechaty instances will set plugins by themselves, and only by themselves.

src/wechaty.ts
```diff
@@ -202,6 +215,38 @@ export class Wechaty extends EventEmitter implements Sayable {
...
+   *     this.options = options;
+   *   }
+   *   install(bot) {
+   *     bot.on('message', message => http.post(this.options.url, { data: message }))
+   *   }
+   * }
+   *
+   * bot.use(new WechatyReportPlugin({ url: 'http://somewhere.to.report.your.data.com' });
+   */
+  public static use (
```
huan, line 240: It's great to be able to set multiple plugins at once.

Instead of `use([])`, how about let's be able to call `use(...plugins)` , so that it will be more concise?

src/wechaty.ts
```diff
@@ -202,6 +215,38 @@ export class Wechaty extends EventEmitter implements Sayable {
...
+    plugins: WechatyPlugin | WechatyPlugin[],
+  ) {
+    ([] as WechatyPlugin[]).concat(plugins).forEach(plugin => {
```
huan, line 243: If it is not a must, I feel that we can just set a plugin and forget that.

So the `globalPlugins` might not a must at here?

As I asked the question about this, I'd like to learn more from you about the reason that you want to use a global class method of the `use()`: If it's necessary, we can keep it; if it can be forgotten, then let's just forget it.

src/wechaty.ts
```diff
@@ -282,6 +327,8 @@ export class Wechaty extends EventEmitter implements Sayable {
     // No need to set puppet/wechaty, so do not clone
     this.UrlLink        = UrlLink
     this.MiniProgram    = MiniProgram
+
+    this.installGloablPlugin()
```
huan, line 331: This will make all the Wechaty instances be affected by the other instances, it's more like to be introducing problems than the benefits here.

src/wechaty.ts
```diff
@@ -521,6 +568,51 @@ export class Wechaty extends EventEmitter implements Sayable {
...
+   *     bot.on('message', message => http.post(this.options.url, { data: message }))
+   *   }
+   * }
+   *
+   * bot.use(new WechatyReportPlugin({ url: 'http://somewhere.to.report.your.data.com' });
+   */
+  public use (plugins: WechatyPlugin | WechatyPlugin[]) {
+    ([] as WechatyPlugin[]).concat(plugins).forEach(plugin => {
+      if (!this.plugins.includes(plugin)) {
+        this.plugins.push(plugin)
```
huan, line 596: The same comment as the `static use()` before: I'd like to be able to _set and forget_ the plugins... if it can be forgotten.

src/wechaty.ts
```diff
@@ -521,6 +568,51 @@ export class Wechaty extends EventEmitter implements Sayable {
...
+        this.installPlugin(plugin)
+      }
+    })
+    return this
+  }
+
+  private installGloablPlugin () {
```
huan, line 603: Looking forward to learning from you that why do we need to install global plugins.

I feel that it can be archived for instance by instance as well.

src/wechaty.ts
```diff
@@ -521,6 +568,51 @@ export class Wechaty extends EventEmitter implements Sayable {
...
+    Wechaty.globalPlugins.forEach(plugin => this.installPlugin(plugin))
+  }
+
+  private installPlugin (plugin: WechatyPlugin) {
+    if (plugin.installed) {
+      return
```
huan, line 609: About install one plugin multiple times: I believe it should be the responsibility of the user who is setting the plugin, to make sure everything is ok.

If we skip one plugin in silence for some reason, I believe it will introduce more puzzles for some use-cases.

src/wechaty.ts
Gcaufy: It's can be forgotten. but I'm just thinking maybe the user wants to know what plugins they have installed.

once we forget it, user can never know what plugins they have installed.

src/wechaty.ts
Gcaufy: if you have two bots, then you have to install for every instance.
in this case, you can just install it as a global plugin.


like what it did in Vue
you can add filter in both `Vue.filter()` or `new Vue({filter: {}})`

src/wechaty.ts
Gcaufy: Functional plugin is fine, if you wish to use.
In my project I also use Functional plugin. and I just feel it make code more unreadable.

In webpack, they use Class style plugin. so that's inspired me. 
No need to understand why I have to return function () {};  and what is the `this`. just need to write a plugin class with a  specific method. and you can also write logic in the constructor, or some private method. if the plugin is complex.

but this is fine, if you prefer Functional style, I can change it back.

src/wechaty.ts
Gcaufy: explained below

src/wechaty.ts
Gcaufy: explained below

src/wechaty.ts
Gcaufy: I don't get it.

src/wechaty.ts
huan: 1. No need to understand why I have to return function () {};
1. write the logic in the constructor or some private method. if the plugin is complex.

I believe the above two reasons are quite strong to support the Class style, and I agree with most of your points.

So please feel free to make your final decision for choosing which style we should use. I'll be happy no matter which style you have chosen: `Class` and `Functional` style are all good.

src/wechaty.ts
huan: Awesome!

src/wechaty.ts
huan: If it can be forgotten, how about leaving the "what plugins have I installed" responsibility to the end-user who is adding plugins?

Because:

1. They can save their plugins to an array for future use very easily (if needed)
1. `Wechaty.plugins` will make one more property in Wechaty class, which there have already a lot. (I always want to move them out to external instead of keeping them internal)

src/wechaty.ts
huan: Got it, and I'm very happy to know that it seems you want to create lots of Wechaty instances and will manage them with your plugins!

If VUE has this kind of design, I agree that we can try this design because that's a lot of front developers who will very familiar with that.

Please go ahead with the `installGlobalPlugin()` and implement it.

src/wechaty.ts
huan: Never mind, I misunderstood your code.

`this.installGlobalPlugin()` is good.

src/wechaty.ts
Gcaufy: ok. will remove the property

src/wechaty.ts
Gcaufy: ok, since we removed the wehcaty.plugins property. we won't know they install twice. it's not a problem any more.

Gcaufy:
Here are the conclusions, just want to make sure I'm not missing anything.

1. Use the name 'Plugin' (keep it)
2. Use both Wehcaty.use and wechaty.use (keep it)
3. Remove the installed plugins property. (need changes)
4. Change plugin to Functional style. (need changes)


btw. I probably need to work this in this weekend.

src/wechaty.ts
```diff
@@ -114,13 +114,8 @@ export interface WechatyOptions {
...
 
-// A plugin is a Class with a install function.
-export abstract class WechatyPlugin {
-
-  abstract install (instance: Wechaty): void
-
-  abstract installed?: boolean
-
+export interface WechatyPlugin {
+  (this: Wechaty): void;
```
huan, line 118: How about let's use an explicit args for passing the `wechaty` instead of using `this`?

I'll write the reason in detail in the comments outside.

huan:
Everything looks good, thank you very much @Gcaufy !

I have one concern about the `this` for our Plugin function, I have left my comment as suggesting to use an explicit `wechaty` arg instead of the `this`, the reason is as the following:

-----

In my experience in the past years, there are lots of developers who do not understand what `this` is in the JS function.

They will be very confused about the `this` and want to know how to use it, like the following screenshot, which was posted in the past week in our developers' home WeChat room:

image

image

I personally like to use `this`, and this is the current design of the event listener callback functions in Wechaty.

However, according to what the above screenshot shows out, I'd like to suggest that we use a `wechaty` arg to archive that, to make it easier to understand for all developers.

src/wechaty.spec.ts
```diff
@@ -149,6 +149,38 @@ test('on(event, Function)', async t => {
...
+      this.on('message', () => (result += 'FROM_GLOBAL_PLUGIN:'))
+    }
+  }
+
+  const myPlugin = function () {
+    return function (this: Wechaty) {
+      this.on('message', () => (result += 'FROM_MY_PLUGIN:'))
+    }
+  }
+  Wechaty.use(myGlobalPlugin())
```
huan, line 166:
```ts
class WechatyTest extends Wechaty {}
WechatyTest.use(myGlobalPlugin())
```

This would be better because it will not modify the global `Wechaty` class.

src/wechaty.ts
```diff
@@ -152,6 +156,8 @@ export class Wechaty extends EventEmitter implements Sayable {
    */
   private static globalInstance: Wechaty
 
+  private static globalPlugins: WechatyPlugin[] = []
```
huan, line 159: To make the variable name style to be consistent, I'd like to suggest to rename the `globalPlugins` to `globalPluginList`.

src/wechaty.spec.ts
Gcaufy: agree

Gcaufy:
> Everything looks good, thank you very much @Gcaufy !
> 
> I have one concern about the `this` for our Plugin function, I have left my comment as suggesting to use an explicit `wechaty` arg instead of the `this`, the reason is as the following:
> 
> In my experience in the past years, there are lots of developers who do not understand what `this` is in the JS function.


I don't like this either. One of the reason to use Class style is to avoid using `this`.

huan:
kindly ping @Gcaufy 

I believe that after we change the `this` to `(wechaty: Wechaty)` and a few variable names, this PR is ready for merge.

I cannot be waiting to use this great new feature!

Gcaufy:
Changes:

1. Renamed:  globalPlugins -> globalPluginList
2. Fix this in pluign
3. Test case updated. use extends Wechaty
4. A bug fix: when use extends, Wechaty.globalPluginList is wrong.

huan:
Thanks to this great Plugin system, it will greatly improve the Wechaty ecosystem by letting everyone writing a powerful plugin to archive their needs with ease!

This feature will be available from Wechaty version 0.39.19

PR #1955, opened 2020-04-28
Friendship contact

su-chang:
## I'm submitting a...

```
[x] Bug Fix
[ ] Feature
```

## Checklist

## Description

```
[ ] Yes
[ ] No
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "wechaty",
-  "version": "0.39.14",
+  "version": "0.39.15",
   "description": "Wechaty is a Bot SDK for Individual Account, Powered by TypeScript, Docker, and 💖",
   "main": "dist/src/index.js",
   "typings": "dist/src/index.d.ts",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "wechaty",
-  "version": "0.39.15",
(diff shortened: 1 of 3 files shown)


su-chang:
Related issue: https://github.com/wechaty/wechaty/issues/1954

huan:
Thanks for the PR!

I have a question: it seems the `ready()` in `contact()` method is not necessary for puppets other than Donut, did you know the reason for that?

BTW: this is a breaking change, I'd like to find the fundamental reason for this problem, and I hope we can fix it without any breaking change.

su-chang:
> I have a question: it seems the `ready()` in `contact()` method is not necessary for puppets other than Donut, did you know the reason for that?

I do not know about that. Could you please explain it for me?

> BTW: this is a breaking change, I'd like to find the fundamental reason for this problem, and I hope we can fix it without any breaking change.

Yeah, it is a breaking change, but I think it is better for user to get contact payload rather than an empty Contact object.

huan:
Instead of introducing a breaking change, I'd like to move `ready()` inside to archive the same goal.

Please see my latest changes in your PR, and let me know what you think.

su-chang:
Understand, it's a better way than what I did! 

Thank you very much.

Cheers!

PR #1968, opened 2020-05-10
Fix async error handler

Gcaufy:
## I'm submitting a...

```
[x] Bug Fix
[ ] Feature
```

## Checklist

- [x] Tests for the changes have been added

## Description

https://github.com/wechaty/wechaty/blob/master/src/wechaty.ts#L618

This code doesn't handle async errors, which means they user have to write code like this, other wise the error will goes to unhandled promises.

```
bot.on('message', (m) => {
  try {
    // do all my works
  } catch(e) {
   // handle all my errors
  }
})
```

```
[x] Yes
[ ] No
```

This is going t break thing. the user's `emit('error')` may get more errors than before.

### DIFF
--- a/src/wechaty.ts
+++ b/src/wechaty.ts
@@ -585,7 +585,7 @@ export class Wechaty extends EventEmitter implements Sayable {
     hotImport(absoluteFilename)
       .then((func: AnyFunction) => super.on(event, (...args: any[]) => {
         try {
-          func.apply(this, args)
+          return func.apply(this, args)
         } catch (e) {
           log.error('Wechaty', 'addListenerModuleFile(%s, %s) listener exception: %s',
             event, modulePath, e,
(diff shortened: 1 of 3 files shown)


src/wechaty.ts
```diff
@@ -610,15 +610,22 @@ export class Wechaty extends EventEmitter implements Sayable {
   private addListenerFunction (event: WechatyEventName, listener: AnyFunction): void {
     log.verbose('Wechaty', 'addListenerFunction(%s)', event)
 
+    const handleError = (e: Error, type = '') => {
+      log.error('Wechaty', 'addListenerFunction(%s) listener %sexception: %s', event, type ? '' : (type + ''), e)
```
windmemory, line 614: Seems like the `type` is a string and will always be a string, so I don't think we need to do `type ? '' : (type + '')` here, which is redundant.

```suggestion
      log.error('Wechaty', 'addListenerFunction(%s) listener %s exception: %s', event, type, e)
```

src/wechaty.ts
```diff
@@ -610,15 +610,22 @@ export class Wechaty extends EventEmitter implements Sayable {
...
+      this.emit('error', e)
+    }
+
     /**
      * We use `super.on()` at here to prevent loop
      */
     super.on(event, (...args: any[]) => {
       try {
-        listener.apply(this, args)
+        const rst = listener.apply(this, args)
```
windmemory, line 623: Is `rst` stands for `result`? If so, I would prefer to use `result` here.

huan:
@Gcaufy  Thank you very much for adding the async error support for Wechaty, I believe it's necessary to catch the async error for our system and emit an `error` event to prevent uncaught promise rejection.

@windmemory Thanks for the code review, please help @Gcaufy to get this PR approved from you, then we will be able to merge it.

Wechaty benefited lots of the efforts from you for the good, I really appreciate it!

src/wechaty.ts
Gcaufy: Have no idea why I did it like this. lol

windmemory:
@huan This PR looks good to me, approved already!

huan:
Thank you very much for enhancing the error handler for our listeners @Gcaufy @windmemory, appreciate it!

PR #1976, opened 2020-05-18
Detail error stack

windmemory:
Refer to https://github.com/juzibot/wechaty-puppet-donut/issues/25#issuecomment-629963339

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "wechaty",
-  "version": "0.39.31",
+  "version": "0.39.32",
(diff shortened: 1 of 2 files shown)


huan:
Thank you very much for improving the Exception/Rejection error message!

However, those messages should always be taken care of by the developers: it should be caught by the local `try {} catch {}` block, and that will be the right way to deal with it.

windmemory:
>However, those messages should always be taken care of by the developers: it should be caught by the local try {} catch {} block, and that will be the right way to deal with it.

Yes, I totally agree with that!

PR #1993, opened 2020-06-09
test: update workflows for node ^12

kis87988:
## I'm submitting a...

```
[ ] Bug Fix
[ ] Feature
[x] Other (Refactoring, Added tests, Documentation, ...)
```

## Checklist

## Description

I see the commit id:608bbd2f2b6594a7725e1cec21ba6c120cacffa0 with require node v12^
seem reasonable to remove v10, and v11 because that is below the require for the node.

```
[ ] Yes
[x] No
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,7 +13,7 @@ jobs:
       - run:
           name: Install Node.js
           command: |
-            curl --silent --location https://rpm.nodesource.com/setup_10.x | sudo bash -
+            curl --silent --location https://rpm.nodesource.com/setup_12.x | sudo bash -
             sudo yum -y install nodejs
 
       - run:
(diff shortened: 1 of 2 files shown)


huan:
Cloud yo please explain the reason why do you remove the Node.js v10 & v11 from the CI?

kis87988:
@huan,

I see the commit id:608bbd2f2b6594a7725e1cec21ba6c120cacffa0 with require node v12^
seem reasonable to remove v10, and v11 because that is below the require for the node.

But we may still check the old version for the node since some puppet still use v10

huan:
Thank you for the explanation, I think we should drop the Node v10 & v11 because the current Node.js LTS version is 12.

Merged.

PR #2138, opened 2021-02-07
pass custom server host to io server

windmemory:
## Checklist

- [x] Tests for the changes have been added
- [x] CI has been passed. (GitHub actions all turns green)
- [x] SLA has been signed

## Description

> please describe the changes that you are making, with the related issue number.
Related to discussion https://github.com/wechaty/wechaty-puppet-service/issues/123

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -119,6 +119,7 @@ export class Config {
   public default = DEFAULT_SETTING
 
   public apihost = process.env.WECHATY_APIHOST || DEFAULT_SETTING.DEFAULT_APIHOST
+  public customizeServerHost = process.env.WECHATY_CUSTOM_SERVER_HOST || ''
 
   public systemPuppetName (): PuppetModuleName {
     return (
--- a/src/config.ts
+++ b/src/config.ts
@@ -119,7 +119,7 @@ export class Config {
   public default = DEFAULT_SETTING
 
   public apihost = process.env.WECHATY_APIHOST || DEFAULT_SETTING.DEFAULT_APIHOST
-  public customizeServerHost = process.env.WECHATY_CUSTOM_SERVER_HOST || ''
+  public serviceIp = process.env.WECHATY_PUPPET_SERVICE_IP || ''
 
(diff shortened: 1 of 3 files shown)


src/config.ts
```diff
@@ -119,6 +119,7 @@ export class Config {
   public default = DEFAULT_SETTING
 
   public apihost = process.env.WECHATY_APIHOST || DEFAULT_SETTING.DEFAULT_APIHOST
+  public customizeServerHost = process.env.WECHATY_CUSTOM_SERVER_HOST || ''
```
huan, line 122: How about let's consider another environment variable name: `WECHATY_PUPPET_SERVICE_HOST`?

It will be a perfect match to the `WECHATY_PUPPET_SERVCIE_PORT` env name.

See:

https://github.com/wechaty/wechaty/blob/040784c94a137ae9da7fbea8095a9b5d2b385e5c/bin/io-client.ts#L77-L78

src/io.ts
```diff
@@ -135,7 +135,7 @@ export class Io {
 
     this.id = options.wechaty.id
 
-    this.protocol = options.protocol + '|' + options.wechaty.id
+    this.protocol = options.protocol + '|' + options.wechaty.id + '|' + config.customizeServerHost + '|' + options.servicePort
```
huan, line 138: Here we can use `options.serviceHost` which is a clearer name I believe.

huan (changes requested):
I have a rename suggestion, if you agree with me please update the name then we can merge it!

windmemory:
Name updated, feels better than the original ones.

huan (approved):
LGTM

huan:
You forgot to bump the version...

windmemory:
@huan You approved too fast... The commit to bump version was being built, because `npm test` is slow, so it finished after you merged the code...

huan:
Aha~ wechaty v0.57.6 is on its way...

And the api.chatie.io has already been updated to the latest version, so please test it and let me know if it works.

Issue #2394, opened 2022-03-30
Feature Request: Message Could Mention All Members in Room but Not Announcement.


su-chang:
## Background
When we want to send message which could mention all members in the room, we could set room announcement for it.

But it will stay on the board until the next new announcement coming. If this message is not so much important to set on the board, we really need some other method to suit this case.

For example, mention all members only. Just like:

`@all hello, everybody here.`

And this feature has been supported in `WeCom` now.

### Design One
Above all, I'd like to make it into method `room.say()` and could transfer a special params like: `[ @all ]`.

**Interface**
```ts
async say (sayable: Sayable | TemplateStringsArray, ...varList: unknown[]): Promise<void | MessageInterface>
```

**Example**
```ts
const room = wechaty.Room.find({ name: 'Mention all members room' })
await room.say('hello, everybody here.', ...[ '@all' ])
```

This design in common sense. But will process with the special params `[ '@all' ]`, and with a little trouble for some puppets adaption.

### Design Two
Another way to support it.

**Interface**
```ts
async announce (text: string, pin: boolean = true) : Promise<void>
```

> Maybe the params `pin` could be replaced by a better name.

**Example**
```ts
const room = wechaty.Room.find({ name: 'Mention all members room' })
await room.announce('hello, everybody here.', false) // would regard it as text message with mention only
```

This design could make use and implement it easily. If you don't want to set an announcement, just transfer a `false` params. It will be a text message which could only mention all members in room.

## Todo List
- [ ] PR in wechaty
- [ ] PR in puppet
- [ ] PR in puppet-service
- [ ] PR in grpc

windmemory:
I would like to vote design one, seems like that one is more common across different IMs.

huan:
I feel design two is more clear to me, and at the same time there will also be no problem across other IMs? (please correct me if I'm wrong @windmemory )

@su-chang Please feel free to go ahead to create a PR to demonstrate your design two, let's continue discussing based on the PR, and thank you very much for your new feature RFC!

windmemory:
I think the main problem for design two is that, the @all symbol could show up in any position inside a message, not only at the beginning, but also somewhere in the middle. So if we do design two, the method will lack of the ability to explicitly set the position of @all.

huan:
Could we just put a string `@all` in the content?

Like:

```ts
room.announce('Attention! @all Please read: ...', false)
```

Does that fulfill your requirement?

windmemory:
[...]

PR #2418, opened 2022-05-17
Feat at all

hcfw007:
#2394

### DIFF
--- a/src/user-modules/room.ts
+++ b/src/user-modules/room.ts
@@ -539,24 +539,7 @@ class RoomMixin extends MixinBase implements SayableSayer {
       /**
        * Mention in the string
        */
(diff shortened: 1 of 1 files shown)


huan:
Could you please also add an unit test so that we can see the behavior of the new method?

Thank you very much!

hcfw007:
> Could you please also add a unit test to see the behavior of the new method?
> 
> Thank you very much!

This PR will require changes in wechaty-puppet first. I'm asking to confirm that this design is fine with you, becuase I find it not very elegant.
```
this.wechaty.puppet.roomAnnounce(
        this.id,
        finalText,
        false,
        metionList.map(c => c.id)
}
```

PR #2533, opened 2023-05-08
Feat/at all

hcfw007:
https://github.com/wechaty/wechaty/issues/2394


huan:
According to my comment on the puppet PR, if we can comfort with it, this PR will be unnecessary anymore.

To-be-discussed.

PR #2550, opened 2023-06-25
fix: 🐛 remove swc dependencies

hcfw007:
https://github.com/wechaty/wechaty/issues/2549

### DIFF
--- a/src/user-mixins/poolify.ts
+++ b/src/user-mixins/poolify.ts
@@ -40,7 +40,7 @@ const poolifyMixin = <MixinBase extends Constructor>(mixinBase: MixinBase) => <T
       return this._pool!  // FIXME: why we need "!" at here?
     }
 
-    public static load<L extends Constructor<T> & PoolifyMixin<T>> (
+    public static load<L extends Constructor<T & {}> & PoolifyMixin<T>> (
       this : L,
       id   : string,
     ): T {
--- a/src/user-mixins/poolify.ts
+++ b/src/user-mixins/poolify.ts
@@ -40,6 +40,11 @@ const poolifyMixin = <MixinBase extends Constructor>(mixinBase: MixinBase) => <T
       return this._pool!  // FIXME: why we need "!" at here?
     }
 
+    /**
+     * Nan(202111): We use it as a workaround to meet the requirement of Constructor parameters
+     *
+     * See: https://github.com/wechaty/wechaty/issues/2553
+     */
     public static load<L extends Constructor<T & {}> & PoolifyMixin<T>> (
       this : L,
       id   : string,
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -3,6 +3,8 @@
   "compilerOptions": {
     "outDir": "dist/esm",
     "lib": ["ESNext", "DOM"],
+    "verbatimModuleSyntax": false,
+    "ignoreDeprecations": "5.0"
   },
   "exclude": [
     "node_modules/",
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -4,6 +4,7 @@
     "outDir": "dist/esm",
     "lib": ["ESNext", "DOM"],
     "verbatimModuleSyntax": false,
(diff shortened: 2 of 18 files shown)


hcfw007:
Why isn't CI running...

tsconfig.json
```diff
@@ -3,6 +3,8 @@
   "compilerOptions": {
     "outDir": "dist/esm",
     "lib": ["ESNext", "DOM"],
+    "verbatimModuleSyntax": false,
```
huan, line 6: Can you explain why we need this? And is it a workaround need to be fixed later, or should it be kept in future versions? So does the next line.

src/user-mixins/poolify.ts
```diff
@@ -40,7 +40,7 @@ const poolifyMixin = <MixinBase extends Constructor>(mixinBase: MixinBase) => <T
       return this._pool!  // FIXME: why we need "!" at here?
     }
 
-    public static load<L extends Constructor<T> & PoolifyMixin<T>> (
+    public static load<L extends Constructor<T & {}> & PoolifyMixin<T>> (
```
huan, line 43: Why do we need to add `& {}` here?

Can you explain in detail?

And does it a workaround and need to be fixed in the future versions, or it should be kept with a reasonable explanation?

tsconfig.json
hcfw007: This is a workaround before @chatie/tsconfig updates. ```verbatimModuleSyntax``` is the replacement of ```importsNotUsedAsValues``` after typescript 5.0. See: https://www.typescriptlang.org/tsconfig#importsNotUsedAsValues

src/user-mixins/poolify.ts
hcfw007: This should be kept because Constructor type now reaquires the first parameter to extend {}.
```ts
 type Constructor<
  I extends {} = {},
  C = {}
> = ClassInterface<C> & InstanceInterface<I>
```

hcfw007: I tried to add <T extends {}> here:
```ts
const poolifyMixin = <MixinBase extends Constructor>(mixinBase: MixinBase) => <T>() => 
```

But it somehow breaks the whole system.

tsconfig.json
huan: If it's a workaround, can you file an issue to record it and put the issue link in the comment so we can track it in the future? Please also do this in similar changes because it's important for future maintenance. Thanks!

src/user-mixins/poolify.ts
huan: This is very tricky because with your modification, your code will allow to pass an `T` does not extends `{}` to be passed to the `Constructor`, which is most likely a workaround.

hcfw007: Yeah the best practice I believe is to make T extend {} in the mixn declaration. But that will break other user-modules. (recursively references).
I don't know the reason for now.

tsconfig.json
hcfw007: https://github.com/wechaty/wechaty/issues/2551

src/user-mixins/poolify.ts
huan: That's why I'm asking whether it's a workaround that should be fixed in the future or if it should be kept as a regular code.

So it looks more like a workaround now. Could you please file an issue to record this and add the issue link to the source code you modified? Because doing so, we can track the workaround and make the maintenance work in the future possible.

Thank you very much!

hcfw007: https://github.com/wechaty/wechaty/issues/2553

huan: Could you please add comments in the code to brief explain and link to the issue?

thank you very much!

hcfw007: Sure and done

huan (approved):
LGTM

Thank you very much for fixing them with the issue explanations!

Issue #2454, opened 2022-11-08
docker部署，扫码登录后报503错误


isbiong100:
WECHATY Environment Variables: 5
WECHATY_TOKEN=“MY_TOCKEN”
WECHATY_DOCKER=1
WECHATY_PUPPET_SERVER_PORT=8080
WECHATY_LOG=verbose
WECHATY_PUPPET=wechaty-puppet-wechat

tarting Docker Container for Wechaty v1.20.2 with Node.js v16.15.0 

报错日志：
09:00:33 INFO IoClient [4] https://login.weixin.qq.com/l/wZ2a6cNrjA==
Online QR Code Image: https://wechaty.js.org/qrcode/https%3A%2F%2Flogin.weixin.qq.com%2Fl%2FwZ2a6cNrjA%3D%3D
09:00:34 VERB PuppetWeChatBridge onLoad() page.url=https://wx.qq.com/cgi-bin/mmwebwx-bin/webwxnewloginpage?ticket=AS8SYJEE1ut7ltDHUltpn81k@qrticket_0&uuid=wZ2a6cNrjA==&lang=en_US&scan=1667898024
09:00:34 VERB PuppetWeChatBridge readyAngular()
09:00:34 VERB PuppetWeChatBridge inject()
09:00:34 VERB PuppetWeChatBridge inject() exception: execute proxyWechaty(init) error: 503, init() without a ready angular env. stack: Error: execute proxyWechaty(init) error: 503, init() without a ready angular env
    at Bridge.inject (file:///wechaty/node_modules/wechaty-puppet-wechat/src/bridge.ts:363:15)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async Bridge.onLoad (file:///wechaty/node_modules/wechaty-puppet-wechat/src/bridge.ts:214:7)
09:00:34 ERR PuppetWeChatBridge onLoad() exception: Error: execute proxyWechaty(init) error: 503, init() without a ready angular env

端口是打开的，docker是没问题的，请问是怎么回事？

ericcug:
你需要自己编译更新使用新版的wechaty-puppet-wechat，官方docker未包含支持uos的puppet。

huan:
A PR request to upgrade the UOS support is welcome!

ericcug:
> > 你需要自己编译更新使用新版的wechaty-puppet-wechat，官方docker未包含支持uos的puppet。
> 
> 请问有具体的方法吗？（虚心求教）

[...]

PR #2555, opened 2023-07-02
fix: fix docker publish workflow.

Shaun-Fong:
# Description

Problems with the current publishing workflow **NETWORK SOCKET TIMEOUT**

image
[Image From](https://github.com/wechaty/wechaty/actions/runs/5415427649/jobs/9843861077)

When I replaced it with [actions-publish-docker@v1.1.2](https://github.com/marketplace/actions/docker-publish) it ran successfully.

This PR should fix this problem.

Refs: #2454

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -32,24 +32,9 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - name: Set up QEMU
-        uses: docker/setup-qemu-action@v1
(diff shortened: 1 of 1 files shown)


huan:
Thanks for creating this PR to fix the Docker build process!

Could you please confirm that the new fix still supports the ARM platform?

We were using QEMU before because we will build our docker image with different platforms, so that we can use Wechaty docker image on a Raspberry Pi.

huan:
Hey @ellipsis, give me a code review

